PR #6, opened 2016-06-15
feature/eager-fetch-tia-comps-on-init

cdaringe:
# problem statement

comps may not all be loaded. only diffs loaded into state on change.
# solution

load 'em all up.


cdaringe:
@swashcap, just 👀

Issue #27, opened 2016-08-23
coinstac-simulator can’t simulate more than 4 clients


swashcap:
**Problem:** _coinstac-simulator_ errors out when the declarations file contains more than 4 users.

Here’s an example error:

``` shell
error: USER loadTestUser2 [36775]: error: Unhandled Rejection at: Promise Promise {
  _bitField: 18087936,
  _fulfillmentHandler0: 
   TypeError: Cannot read property '_id' of undefined
       at consortiaDB.all.then.then.then.then (/Users/creed/coinstac/packages/coinstac-simulator/src/boot-client.js:88:41)
       at runCallback (timers.js:570:20)
       at tryOnImmediate (timers.js:550:5)
       at processImmediate [as _immediateCallback] (timers.js:529:5)
   From previous event:
       at kickoff (/Users/creed/coinstac/packages/coinstac-simulator/src/boot-client.js:78:4)
       at process.on (/Users/creed/coinstac/packages/coinstac-simulator/src/boot-client.js:132:5)
       at emitTwo (events.js:106:13)
       at process.emit (events.js:191:7)
       at process.nextTick (internal/child_process.js:719:12)
       at _combinedTickCallback (internal/process/next_tick.js:67:7)
       at process._tickCallback (internal/process/next_tick.js:98:9),
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined,
  _trace: 
   { Error
       at Promise.longStackTracesCaptureStackTrace [as _captureStackTrace] (/Users/creed/coinstac/packages/coinstac-common/node_modules/bluebird/js/release/debuggability.js:369:19)
       at Promise._then (/Users/creed/coinstac/packages/coinstac-common/node_modules/bluebird/js/release/promise.js:230:17)
       at Promise.then (/Users/creed/coinstac/packages/coinstac-common/node_modules/bluebird/js/release/promise.js:123:17)
       at kickoff (/Users/creed/coinstac/packages/coinstac-simulator/src/boot-client.js:123:4)
       at process.on (/Users/creed/coinstac/packages/coinstac-simulator/src/boot-client.js:132:5)
       at emitTwo (events.js:106:13)
       at process.emit (events.js:191:7)
       at process.nextTick (internal/child_process.js:719:12)
       at _combinedTickCallback (internal/process/next_tick.js:67:7)
       at process._tickCallback (internal/process/next_tick.js:98:9)
[...]

PR #28, opened 2016-08-29
Simulator Cleanup

swashcap:
- **Problem:** _coinstac-simulator_ can only run ~4 clients (see #27 for details). This appears to be due to a race conditioning where client processes can’t get an active consortium document from the database.
- **Solution:** Make client processes retry consortium database requests 4 times to avoid race condition (see 60b893576cb6b9fcf6d8ece4be7dd2926dd9d892).
- **Bonus:** This refactors some simulator code:
  - Move database server booting from a spawned process to into main process. It’s using pouchdb-server, which already spawns a new child process.
  - Makes configuration sensible.
  - Moves CLI to _bin_ directory (should we canonize this for all projects?)
- **Testing:**
  1. Check out the _bug/#27-simulator-clients_ branch
  2. Clone [swashcap/coinstac-simulator-load-tester](https://github.com/swashcap/coinstac-simulator-load-tester)
  3. Link _coinstac-simulator_ globally and link it to _coinstac-simulator-load-tester_.
  4. Run `node src/index.js -c 10` inside the _-load-tester_ to simulate 10 clients.


packages/coinstac-simulator/src/db-server.js
```diff
@@ -5,58 +5,126 @@
...
 
-require('./handle-errors')();
-const pdbs = require('spawn-pouchdb-server');
-const cp = require('child_process');
-const pdbsConfig = require('./.pouchdb-server-config');
+require('./utils/handle-errors');
+
 const cloneDeep = require('lodash/cloneDeep');
+const common = require('coinstac-common');
+const config = require('./utils/config');
```
rssk, line 12: convict overkill?

packages/coinstac-simulator/config/default.json
```diff
@@ -0,0 +1,13 @@
+{
+  "backend": false,
```
rssk: Chris suggests this may be more clear under a `pouch-db-server:` sub option

packages/coinstac-simulator/src/db-server.js
swashcap: Eh. I was attempting to log pdbs's output to a file. It was nice to throw it in a _config/local.json_ file. Think that justifies convict-style configuration?

packages/coinstac-simulator/config/default.json
swashcap: Definitely thought about this when making the configuration. We’re only using it for pdbs right now, so I left it at the root. I’ll go back and adjust.

cdaringe: FTR ☝️, @rssk _suggested_ the sub-option.  i drew the conclusion that `coinstac-simulator/config/default.json` should configure `coinstac-simulator`, and hence felt it was misleading if it was actually configuring PDBS

packages/coinstac-simulator/src/boot-clients.js
```diff
@@ -1,48 +1,77 @@
...
-require('./handle-errors')();
 const cp = require('child_process');
-const chalk = require('chalk');
+const logger = require('./utils/logger');
 const path = require('path');
-const logger = require('./logger');
-const userChatOK = chalk.magenta;
 
-module.exports = function bootClients(declPath) {
+/**
```
rssk, line 9: If you dont mind, and you understand these funcs well, mind writing some more explanation in the docblocs for future generations?

packages/coinstac-simulator/src/db-server.js
```diff
@@ -5,58 +5,126 @@
...
+        reject(error);
+      } else {
+        server = srv;
+        resolve(server);
+      }
     });
-  });
-};
+  })
+    .then(() => {
```
rssk, line 71: wat indent

packages/coinstac-simulator/src/db-server.js
rssk: Second thought, I think its fine as is, we can revisit if more stuff gets tossed in

packages/coinstac-simulator/src/db-server.js
```diff
@@ -5,58 +5,126 @@
...
+      return server;
+    });
+}
+
+/**
+ * @returns {Promise}
+ */
+function teardown() {
+  if (!server) {
+    return Promise.reject(new Error('Server not instantiated'));
```
rssk, line 112: was this an issue?

packages/coinstac-simulator/src/index.js
```diff
@@ -99,7 +99,8 @@ const exportList = {
 
 // Ensure all processes are killed
 process.on('exit', () => {
-  flatten(values(processes)).forEach(p => p.kill());
+  dbServer.teardown().catch(noop);
```
rssk, line 102: 👍

packages/coinstac-simulator/src/utils/handle-errors.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+require('trace');
+require('clarity');
+
+Error.stackTraceLimit = Infinity;
```
rssk, line 6: to infinity, and beyond 💥

packages/coinstac-simulator/src/boot-clients.js
swashcap: A future PR has docblocks on these fellas.

packages/coinstac-simulator/src/db-server.js
swashcap: Looks fine to me. Without confusing diff: https://github.com/MRN-Code/coinstac/blob/934ad12193ae53a6cd3c376a264bffefa8db73dc/packages/coinstac-simulator/src/db-server.js#L70

packages/coinstac-simulator/src/db-server.js
swashcap: It was because I was being a dummy in _src/index.js_. Although it makes sense: can’t tear down the server it isn’t yet created.

packages/coinstac-simulator/config/default.json
swashcap: Updated in 67fb7ef16a8dd9f0306ff08dd99edb51a167718c.

Issue #23, opened 2016-08-19
Improve Algorithm Simulator API


swashcap:
As per the Albuquerque meeting on 2016-08-15 and 2016-08-16 and the discussion with Sergey on 2016-08-17, we need to improve _coinstac-simulator_ so that algorithm authors can start using it.

_coinstac-simulator_ already functions as an executable, and Sergey already has it installed globally. Great! I think this method provides the least setup friction for authors. Sergey has a _declaration.js_ file that contains an object that _coinstac-simulator_ consumes. It looks approximately like:

``` js
const n = require('numeric');

module.exports = {
  user1: {
    name: 'user 1',
    data: n.transpose([
      [0.001, 0.002, 0.003],
      // Lots of data here...
    ]),
  },
  user2: {
    name: 'user 2',
    data: n.transpose([
      // More data!
    ]),
  },
  user3: {
    name: 'user 3',
    data: n.transpose([
      // Tons of arrays!
    ]),
  },
  server: {
    // Some opts for remote
  },
};
```

Sergey doesn’t want to continually copy-paste numbers into his _declaration.js_ file.

Here’s a CLI suggestion:

``` shell
coinstac-simulator --declaration declaration.js \
  --user1 path/to/files1/**/*.txt \
  --user2 path/to/files2/**/*.txt \
  --user3 path/to/files3/**/*.txt
```

This would use *NIX globbing to pass target files to the declaration, which would then run. However, this doesn’t allow for user configuration or server vars. Another take:

``` js
const cstacSimulator = require('coinstac-simulator');

module.exports = {
  user1: {
    x: cstacSimulator.loadFiles('./data/**/*.txt'),
    y: cstacSimulator.loadFile('./data/path/my-file.txt'),
    z: 3e-3,
    aa: [0.011, 0.015, 0.513, 0.744],
  },
  user2: {
    // ...
  },
};
```

If we export utilities for file loading on on _coinstac-simulator_ it could help authors out. Alternatively:
- We tell them to learn Node.js. Might be too steep a learning curve
- We give code examples in some sort of documentation that shows how to load files. Something like:
  
  ``` js
  // Here's how to get some files into your program!
  const fs = require('fs');
  const glob = require('glob'); // External package for *NIX-like globbing
  const path = require('path'); // Built-in used to resolve paths
  
  // `files` is an array of full file paths
  const files = glob.sync(path.join(__dirname, 'my/data/**/*.txt'));
  
[...]

PR #32, opened 2016-08-31
Add File Loading Utility per #23

rssk:
- **Problems:**
  - No easy way to load files into declarations
  - Main issues laid out in #23 
- **Solution:** 
  Create file-loader to load files and turn a known directory structure into an array of user objects. File loader can load a single group of files, either a text delimited or a json array, that it will concat together, or you can point it to a directory structure and/or a structured CSV and it will create the user objects for you.
- **Testing:** `npm test`


packages/coinstac-simulator/src/file-loader.js
```diff
@@ -0,0 +1,110 @@
...
+  const delimiter = delim || ',';
+  const files = glob.sync(fileGlob);
+  let userData = [];
+  files.forEach((filePath) => {
+    if (fs.lstatSync(filePath).isFile()) {
+      const data = fs.readFileSync(filePath, 'utf8');
+      const ext = path.extname(filePath);
+      if (ext === '.json') {
+        // expects JSON array
+        userData = userData.concat(JSON.parse(data));
```
rssk, line 26: This all just concats the arrays together for one 'variable', it might? be nice to have a 'mode' option to do other operations on multiple files passed into one variable. However it may not be a usual case to have multiple files per var, unsure.

packages/coinstac-simulator/src/file-loader.js
```diff
@@ -0,0 +1,110 @@
...
+      const pathInfo = file.replace(baseDir, '').split('/');
+      // assumes the first glob dir is the users
+      // and second is the variable
+      if (!userData[pathInfo[0]]) {
+        userData[pathInfo[0]] = { [pathInfo[1]]: [] };
+      } else if (!userData[pathInfo[0]][pathInfo[1]]) {
+        userData[pathInfo[0]][pathInfo[1]] = [];
+      }
+      userData[pathInfo[0]][pathInfo[1]] = userData[pathInfo[0]][pathInfo[1]]
+        .concat(loadFiles(file, delimiter));
```
rssk, line 90: similar issue to above, just concats

packages/coinstac-simulator/src/file-loader.js
```diff
@@ -0,0 +1,110 @@
...
+      } else {
+        if (delimiter.match(/\n|\r/g)) {
+          userData = userData.concat(data.replace(/\n$|\r$/, '').split(delimiter));
+        } else {
+          userData = userData.concat(data.replace(/\n|\r/g, '').split(delimiter));
+        }
+      }
+    }
+  });
+  return userData;
```
rssk, line 36: This also does no casting, so if you have an 'csv' array, it'll turn up as an array of strings vs a json array will correctly be integers/floats. I think we should cast, but I'd like feedback

packages/coinstac-simulator/src/file-loader.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+  }
+  // convert our object to an array
+  return Object.keys(userData).map((key) => { return userData[key]; });
+};
+
+module.exports = {
+  createUserData,
+  loadFiles,
+  loadFile: loadFiles,
```
swashcap, line 109: :+1: Nice

packages/coinstac-simulator/src/file-loader.js
swashcap: This is fine for now I think. We'll have to get it in the hands of algo authors and see what they want.

packages/coinstac-simulator/src/index.js
```diff
@@ -99,4 +100,7 @@ module.exports = {
       });
     });
   },
+
+  // load in file loading utlilities
+  fileLoader,
```
swashcap: What do you think of exposing all of the file loading methods on the root export? This seems easier:

``` js
const cstacSim = require('coinstac-simulator');

module.exports = {
  local: [{
    x: cstacSim.loadFiles('./path/to/sweet/*.csv'),
    y: cstacSim.loadFile('./path/to/other/*.stuff'),
  }],
};
```

As opposed to calling out `cstacSim.fileLoader.loadFiles` or whatever.

rssk: Agreed 👍

swashcap:
🎉 Great work!

cdaringe:
This is some seriously sexy JavaScript!

rssk:
😊

PR #39, opened 2016-09-13
Fully resolve a declaration's computation path.

### LINKED ISSUE #37: Declaration’s computation path resolution
Computation paths don’t resolve properly:

``` js
module.exports = {
  computationPath: './path/to/thingy.js',
}
```

The above doesn’t resolve. Using something like:

``` js
const path = require('path');

module.exports = {
  computationPath: path.resolve(__dirname, './path/to/thingy.js'),
}
```

…works.

swashcap:
- **Problem:** A declaration’s `computationPath` property isn’t always resolved to the computation file’s location.
- **Solution:** Use Node.js’s built-in module and path resolution to _deal wit it_.

This closes #37.

### DIFF
--- a/packages/coinstac-simulator/src/index.js
+++ b/packages/coinstac-simulator/src/index.js
@@ -97,7 +97,10 @@ const exportList = {
           undefined;
 
         return Promise.all([
-          declaration.computationPath,
+          path.resolve(
+            path.dirname(require.resolve(declarationPath)),
+            declaration.computationPath
+          ),
           local,
           remote,
           !!declaration.verbose,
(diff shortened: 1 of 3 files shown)


packages/coinstac-simulator/src/index.js
```diff
@@ -97,7 +97,10 @@ const exportList = {
           undefined;
 
         return Promise.all([
-          declaration.computationPath,
+          path.resolve(
+            path.dirname(require.resolve(declarationPath)),
```
rssk, line 101: Interesting, never knew that was a thing

PR #40, opened 2016-09-13
Ensure command computations set CWD relative to computation file.

### LINKED ISSUE #38: Simulator miss-sets CWD for CommandComputation
_coinstac-simulator_ doesn’t properly resolve the `cwd` for `CommandComputation`s.

swashcap:
This refactors pool-initializer.js and stub-computation-to-registry.js into the
new get-pool-config.js. This file now creates the DBRegistry and
ComputationRegistry instances necessary for a PipelineRunnerPool and configures
the ComputationRegistry to include the computation under simulation. It also
addresses the computation registry's path resolution issue, which fixes the
command computation's bad CWD problem.

Closes #38.


packages/coinstac-simulator/src/boot-client.js
```diff
@@ -67,16 +69,6 @@ function boot({
       pool.events.on('error', logger.error);
 
       return pool.init();
-    })
-    .then(() => {
```
swashcap, line 71: `getPoolConfig` takes care of all this in both the client and server processes.

packages/coinstac-simulator/src/boot-client.js
```diff
@@ -201,7 +193,7 @@ const kickoff = function kickoff() {
 process.on('message', (opts) => {
   if (opts.boot) {
     boot(opts.boot)
-    .then((result) => process.send({ ready: true, result }));
```
swashcap, line 204: `result` is now the return value of `PipelineRunnerPool#init`, which contains self-references. This causes `process.send` to crash.

packages/coinstac-simulator/test/coinstac-simulator.js
```diff
@@ -375,3 +375,12 @@ tape('runs declaration :: errors', t => {
     });
 });
 
+tape('sets commands\' cwd', t => {
```
swashcap, line 378: You betcha, we testin’.

packages/coinstac-simulator/test/mocks/exec-computation.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = {
```
swashcap, line 3: Hard to tell whether this totally covers us, but `CommandComputation`s in both `local` and `remote` should be good. Also, this test is _noisy_.

rssk (approved):
Looks good 💭

packages/coinstac-simulator/src/boot-client.js
rssk: Yo dawg, I heard you like pools....

swashcap:
@rssk How do you feel about merging?

PR #55, opened 2016-09-29
Address Connection Error

### LINKED ISSUE #41: Connection Error/Timeout
Connection errors and timeouts cause the client to explode. This is due to logging of a PouchDB error with a circular reference.

swashcap:
- **Problem:** PouchDB internal errors contain a circular reference, which causes error logging to blow up due to exceeding max call stack.
- **Solution (kinda):** Update PouchDB? I can’t easily reproduce the error, although pouchdb/pouchdb#5101 looks like it might be the issue. We’re passing around RPC’d objects _everywhere_.

Closes #41.


rssk:
Hey, getting to these tonight, should we not lock down these ver #s? Looks like there are new bug patches:
`npm WARN pouchdb-adapter-memory@6.0.7 requires a peer of pouchdb-core@6.0.7 but none was installed.
npm WARN pouchdb-adapter-leveldb-core@6.0.7 requires a peer of pouchdb-core@6.0.7 but none was installed.
npm WARN pouchdb-adapter-utils@6.0.7 requires a peer of pouchdb-core@6.0.7 but none was installed.`

PR #71, opened 2016-10-17
Bundle Computations

swashcap:
- **Problem:** Installing computations at execution time results in built application crashes ☠️ :boom:
- **Solution:** Make computations npm dependencies, strip unneeded functionality from `ComputationRegistry`. This has the added benefit of ensuring GitHub API limits don't break COINSTAC functionality.
- **Testing:**
  - Run monorepo tests
  - Build application and run a computation


swashcap:
Turns out we _can not_ asar (compress) the build and run executables (see [swashcap/electron-packed-executable](https://github.com/swashcap/electron-packed-executable)). This is unfortunately limiting.

packages/coinstac-common/src/services/classes/computation-registry.js
```diff
@@ -417,44 +208,10 @@ class ComputationRegistry {
     }
     return null;
   }
-
```
rssk, line 420: 👋

PR #72, opened 2016-10-18
User data and computation state clearing

### LINKED ISSUE #59: State clearing via the UI
Add a button in the UI that clears local state. It should have some sort of warning indicating to users that the action deletes all projects.

swashcap:
- **Problem:** Clearing persisted data (on disk) requires running npm scripts. It isn't easy!
- **Solution:** Provide a settings page and buttons to clear the data.
- **Testing:**
  1. Boot _coinstac-ui_ in development mode
  2. Log in
  3. Observe stored data in `$HOME/.coinstac/<username>/`
  4. Go to the settings page
  5. Click the “Delete user data” button
  6. Observe stored data is removed

Closes #59.


swashcap:
UI updates per e5d362ac4d96b0ba1695338bfad3ad95c5ea6041:

delete-em

swashcap:
Computation deletion dropped in 83fd22391b16382c797f26d705f544a68473e648:

drop-comps

packages/coinstac-ui/app/main/services/clean.js
```diff
@@ -0,0 +1,53 @@
...
+        return rimrafAsync(dir).then(() => app.core.teardown());
+      }
+    });
+}
+
+module.exports = {
+  computations: function cleanComputations() {
+    return warnAndDelete(
+      {
+        detail: 'Computations can be re-downloaded',
```
rssk: Does this make sense now we've bundled things?

packages/coinstac-ui/app/main/services/clean.js
```diff
@@ -0,0 +1,53 @@
...
+        message: 'Delete saved computations',
+      },
+      app.core.getComputationsDirectory()
+    );
+  },
+
+  userData: function cleanUserData(username) {
+    return warnAndDelete(
+      {
+        detail: 'You can not undo this action',
```
rssk, line 28: I think we should mention this deletes projects, as its probably not clear what user data is to the enduser

packages/coinstac-ui/app/main/services/clean.js
swashcap: Good point. I'll take it out.

packages/coinstac-ui/app/main/services/clean.js
swashcap: Cool, I’ll update this. I think clarification belongs in the Settings UI, not the alert.

packages/coinstac-ui/app/main/services/clean.js
swashcap: Dropped in a7625dbd64ae28ef7ac8e3b086c56acbd906b146.

packages/coinstac-ui/app/main/services/clean.js
swashcap: Updated in 34764acb4cbab9b1c9254f99d2963c9c64c2cb8d.

PR #86, opened 2016-10-27
Speedup: remove hist addition to speedup runs

rssk:
# Problem

Coinstac is slow, this seems to get worse as computations progress
# Solution

Remove the history from being saved to the database, it's not entirely clear who is responsible for this being slow


swashcap:
Looks like caca3b8bd483d1e0c948fd752258a8af6328b4ff breaks a _coinstac-common_ test. Want to fix it?

cdaringe:
LATER HIST!

PR #93, opened 2016-11-01
Update build scripts to zip, use archiver lib

rssk:
# Problem
tar.gz archives are a bit more difficult for windows users to unpack

# Solution
Zipit if its windows :-)

### DIFF
--- a/packages/coinstac-ui/scripts/build.js
+++ b/packages/coinstac-ui/scripts/build.js
@@ -1,6 +1,6 @@
 const prom = require('bluebird').promisify;
 const packager = prom(require('electron-packager'));
-const zip = require('tar.gz');
+const archiver = require('archiver');
 const fs = require('fs');
 const os = require('os');
 const buildNative = require('./utils/build-native.js');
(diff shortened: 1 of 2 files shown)


packages/coinstac-ui/scripts/build.js
```diff
@@ -22,14 +22,24 @@ buildNative()
...
-  const write = fs.createWriteStream(`${dir}.tar.gz`);
+  console.log(`Finished building at: ${appPath}`); // eslint-disable-line no-console
+  console.log('Now archiving...'); // eslint-disable-line no-console
+
+  if (os.platform() === 'win32') {
+    zip = archiver.create('zip');
+  } else {
+    zip = archiver.create('tar', { gzip: true });
+  }
+  const write = fs.createWriteStream(os.platform() === 'win32' ? `${dir}.zip` : `${dir}.tar.gz`);
```
swashcap: Minor issue, but you’re branching on `os.platform()` twice. Want to move that all into one spot? Or, write `zip` declaration as:

``` js
const zip = os.platform() === 'win32' ? archiver.create('zip') : archiver.create('tar', { gzip: true });
```

Would make it a little easier to read.

swashcap:
:+1:

PR #95, opened 2016-11-03
fix logging issues as well as zipping issues

### LINKED ISSUE #94: Errors are not being logged correctly
# Problem
Module error:
```c:\Users\stac\Desktop\coinstac\packages\coinstac-ui\app\common\utils\log-unhandled-error.js:87
    throw err;
    ^

Error: Cannot find module 'leveldown'```
   
does not log out to file, _all_ errors should log, fix logging logic

### PR DESCRIPTION
**Pull request by rssk**
# Problem
Errors are not being logged to the correct locations and zip files are not valid.

# Solution
Change log location format to always append to the OS specific home directories in both locations where logging happens. Zips were not being passed the correct options as well as the `finalize` method needed to be called.

Closes #94

### DIFF
--- a/packages/coinstac-ui/app/config.js
+++ b/packages/coinstac-ui/app/config.js
@@ -33,7 +33,7 @@ const conf = convict({
     noURLPrefix: true,
   },
   logFile: 'coinstac-log.json',
-  // these are appended to the home dir for you OS 
+  // these are appended to the home dir for you OS
   // *nix: ~/.config/coinstac
   // win: C:\Users\username\AppData\Local\Temp\coinstac
   logLocations: {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/coinstac-ui/app/config.js [hunk 1]
```diff
@@ -33,12 +33,15 @@ const conf = convict({
...
-    darwin: '~/Library/Logs/coinstac/',
-    freebsd: '~/.config/coinstac/',
-    linux: '~/.config/coinstac/',
-    sunos: '~/.config/coinstac/',
-    win32: '$HOME/AppData/Roaming/coinstac/',
+    darwin: 'Library/Logs/coinstac/',
+    freebsd: '.config/coinstac/',
+    linux: '.config/coinstac/',
+    sunos: '.config/coinstac/',
+    win32: 'coinstac/',
```
cdaringe, line 44: would be cool if something like [this](https://www.npmjs.com/package/appdirectory) were more robust.  there may be.  i didn't search very hard.

swashcap:
:+1:

PR #98, opened 2016-11-13
Fix sign up UI

swashcap:
* **Problem:** Signing up via the UI doesn't work
* **Solution:**
  * Ensure _coinstac-client-core_ authenticates newly created users
  * Ensure user is placed in app’s render process’s state upon signup
* **Testing:**
  1. Build a client from this branch:
    ```shell
    git fetch --all
    git checkout bug/#82-account-creation
    cd packages/coinstac-ui
    npm run build
    ```
  2. Clean local state: `npm run clean:db`
  3. Boot the client and attempt to make a new account

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/173590139517418/211870140170035)

### DIFF
--- a/packages/coinstac-client-core/src/index.js
+++ b/packages/coinstac-client-core/src/index.js
@@ -225,10 +225,11 @@ class CoinstacClient {
    * @returns {Promise}
    */
   _initAuthorization(credentials) {
-    if (credentials.email && credentials.name) {
-      return this.auth.createUser(credentials);
-    }
-    return this.auth.login(credentials);
+    const doLogin = () => this.auth.login(credentials);
+
+    return credentials.email && credentials.name ?
+      this.auth.createUser(credentials).then(doLogin) :
+      doLogin();
   }
 
   /**
--- a/packages/coinstac-client-core/src/sub-api/auth.js
+++ b/packages/coinstac-client-core/src/sub-api/auth.js
@@ -46,9 +46,9 @@ class Auth {
     return this.halfpenny.users.post({
       email,
       label: name,  // the API requires a `label` key instead of `name`
-      password,
+      password: btoa(password),
       siteId: '7',
-      username,
+      username: btoa(username),
     });
   }
 
--- a/packages/coinstac-client-core/src/sub-api/auth.js
+++ b/packages/coinstac-client-core/src/sub-api/auth.js
@@ -1,10 +1,8 @@
 'use strict';
 
-const common = require('coinstac-common');
-const User = common.models.User;
 const atob = require('atob');
 const btoa = require('btoa');
-const bluebird = require('bluebird');
+const User = require('coinstac-common').models.User;
 
 const USER_KEY = btoa('COINSTAC_USER');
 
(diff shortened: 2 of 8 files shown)


packages/coinstac-client-core/src/index.js
```diff
@@ -225,10 +225,11 @@ class CoinstacClient {
...
    */
   _initAuthorization(credentials) {
-    if (credentials.email && credentials.name) {
-      return this.auth.createUser(credentials);
-    }
-    return this.auth.login(credentials);
+    const doLogin = () => this.auth.login(credentials);
+
+    return credentials.email && credentials.name ?
+      this.auth.createUser(credentials).then(doLogin) :
```
swashcap, line 231: Big change in _coinstac-client-core_ is here: create that user, then do that login.

packages/coinstac-client-core/src/sub-api/auth.js
```diff
@@ -70,34 +64,27 @@ class Auth {
...
-      if (err.status === 500) {
-        throw new Error(
-          'unable to log into COINS.  Please try again or file a support ticket'
-        );
-      } else if (err.status === 401) {
-        throw new Error('Invalid username or password.  Please try again');
-      }
-      throw err;
-    };
+  login({ password, username }) {
```
swashcap, line 74: Cleanup of general grossness.

packages/coinstac-client-core/test/coinstac-client.js
```diff
@@ -213,3 +213,33 @@ test('CoinstacClient - teardown process', (t) => {
     .catch(t.end);
 });
 
+test('CoinstacClient - auth initialization', t => {
```
swashcap, line 216: Ensures login is called when creating a user. Not the greatest test, I think, but it works.

packages/coinstac-ui/app/main/utils/boot/parse-cli-input.js
```diff
@@ -27,7 +27,7 @@ function parse() {
         data = data.toString();
         // don't fill the screen with useful build information. scan for it,
         // summarize it instead
-        const builtRegex = /.*(\[(built|not cacheable)\]|\{0\}).*[\r\n]/gm;
+        const builtRegex = /.*(\[(built|not cacheable)]|\{0\}).*[\r\n]/gm;
```
swashcap, line 30: I guess that `\` isn’t necessary when the first `[` is escaped? Linter barked at it.

packages/coinstac-ui/app/render/components/form-signup-controller.js
```diff
@@ -43,12 +43,23 @@ class FormSignupController extends Component {
...
-    }));
+
+    if (error) {
+      return FormSignupController.handleSignupError(error);
+    }
+
+    return dispatch(signUp(formData))
+      .then(() => {
+        app.notifications.push({
+          autoDismiss: 2,
```
swashcap, line 54: Notification goes away after 2 seconds. 👋😬 We should make this the default.

packages/coinstac-ui/app/render/state/ducks/auth.js
```diff
@@ -24,13 +24,13 @@ export const logout = applyAsyncLoading(() => {
   };
 });
 
-export const signUp = applyAsyncLoading(user => {
-  return (dispatch) => { // eslint-disable-line
-    return app.core.initialize(user)
-    .then(() => app.notify('success', 'New user created successfully'))
```
swashcap, line 30: Moved notification calls to the component. This makes things clearer: folks look at the component to see the control flow, everything in _app/render/state/_ should deal with **state only**.

packages/coinstac-ui/app/render/state/ducks/util.js
```diff
@@ -1,6 +1,7 @@
 import app from 'ampersand-app';
 
 export const notifyAndThrow = (err) => {
+  app.logger.error(err);
```
swashcap, line 4: Make this slightly less useless. Log it.

packages/coinstac-ui/app/render/components/form-signup-controller.js
rssk: Gah, totally

packages/coinstac-client-core/src/sub-api/auth.js
```diff
@@ -35,27 +33,23 @@ class Auth {
...
-    // @TODO use actual user siteId
-    /* istanbul ignore next */
-    return bluebird.resolve(this.halfpenny.users.post(newUser));
+  createUser({ email, name, password, username }) {
+    return this.halfpenny.users.post({
+      email,
+      label: name,  // the API requires a `label` key instead of `name`
+      password,
+      siteId: '7',
+      username,
```
rssk: username and pass need to be base64 :-) https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/src/lib/controllers/users.js#L144

This is newish for the PR site

swashcap: Fixed in d50e0c74a4e0835b5ac9ecbb978f26a891536384.

PR #99, opened 2016-11-18
Feature/common tidy

cdaringe:
# problem statements

- lots of unused deps are declared in common
- i want to use `Base` in a different project knowing the `joi` i use is the same that `joi` that common will use

# solution

- tidy deps
- expose `joi`


swashcap:
:+1: (This actually passed CI on a bare commit.)

PR #123, opened 2017-01-14
Computation selection

swashcap:
This adds support for algorithm metadata. Authors should add it to their computations’ _package.json_ under the `coinstac` key. The supported properties are `name`, `description`, and `tags`:

```js
{
  "name": "decentralized-single-shot-ridge-regression",
  "version": "1.1.0",
  // ...
  "coinstac": {
    "name": "Single-Shot Ridge Regression",
    "description": "Decentralized ridge regression algorithm...",
    "tags": ["regression"]
  }
}
```

These values are persisted to the _computations_ database by _coinstac-server-core_ and relayed to the user interface:

comp-selector

Single-shot’s the new default.


swashcap:
Closeup of that UI:

closeup

Still need to add some categorization, and we’ll have to do something about >6 computations – maybe a scrollbar for the `div`?

swashcap:
Dropped the Lerna upgrade (to beta.32) from this PR as their new symlink-ing strategy doesn’t play well with our code.

709b01d56bb6060feadbaa085c9fe8742b87f9c0 adds a hard-coded “Regressions” label to the computation selection UI:

do-it-do-it

We can revisit computation tagging/categorization in the future. 👏

Issue #101, opened 2016-11-23
Project’s CSV → Covariate mapping not cleared


swashcap:
A user’s project’s CSV to covariate mapping isn’t cleared when the CSV file is changed. This property should be cleared on the project model because the data underlying the mapping is (probably) different.

PR #125, opened 2017-01-17
Clear covariate mapping when metadata file is removed.

swashcap:
* **Problem:** See #101.
* **Solution:** Reset `metaCovariateMapping` and `metaFile` project state props in the metafile removal handler.
* **Testing:**
  1. Start up _coinstac-ui_
  2. Create a consortium, ensuring covariates are added
  3. Create a new project
  4. Add a metafile to the new project
  5. Map the consortium’s covariates to the metafile
  6. Remove and re-add the metafile
  7. Observe the covariate mappings are reset


swashcap:
Confirmed this worked. Merging it!

PR #126, opened 2017-01-18
Computation Output

swashcap:
* **Problem:** Computation output isn’t useful for researchers
* **Solution:** Add tables to computation results
* **Related:**  _(Merge these first and bump their package.json versions!)_
  * MRN-Code/decentralized-laplacian-ridge-regression#10
  * MRN-Code/decentralized-single-shot-ridge-regression#11


swashcap:
#131 merged rebased versions of these refs. Closing.

PR #127, opened 2017-01-23
Refactor test script

swashcap:
* **Problem:** Repo-wide test script doesn’t run tests in every package. Also, tests state collide (PouchDB server?) causing strange errors.
* **Solution:** Re-write tests to run in series. Test _all the packages!_
* **Testing:** `npm test` from the repo root.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
-  "lerna": "2.0.0-beta.30",
+  "lerna": "2.0.0-beta.32",
   "version": "2.2.2"
 }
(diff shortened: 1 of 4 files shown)


swashcap:
Also: bumping Lerna. Lerna has trouble installing NPM modules that are GitHub deps in _beta.30_. Driving me bananas! 🍌🍌🍌

swashcap:
Also _also_: 6be91bae6db4aa9011db4148a14af4ae633548f7 ensures that `build-native` only builds the root leveldown in _packages/coinstac-ui_ by adding a `--depth=0` flag to the `npm rebuild` command. This is a hack that relies on npm flat-packing dependencies, but ¯\\\_(ツ)_/¯

PR #128, opened 2017-01-25
Pipeline Cleanup

swashcap:
* **Problem:** _coinstac-common_’s `Pipeline` class is difficult to reason about.
* **Solution:** Improve `Pipeline` methods, documentation.
* **Testing:** `npm test` from the monorepo root.

### DIFF
--- a/packages/coinstac-common/src/models/pipeline/pipeline.js
+++ b/packages/coinstac-common/src/models/pipeline/pipeline.js
@@ -141,7 +141,7 @@ class Pipeline extends Base {
       },
     };
 
-    return this.tryNext(runInputs, { preRun: true })
+    return this.maybeIncrementStep(runInputs, true)
       .then(() => {
         this.events.emit('computation:start', runInputs);
         return this._preRunPlugins({ runInputs, compResult, pluginHooks });
@@ -198,17 +198,17 @@ class Pipeline extends Base {
     this.events.emit('computation:end', runInputs);
     const postRunInputs = assign({}, runInputs, { previousData: runOutput });
     return (runCancelled ?
-      Promise.resolve({ didStep: false }) :
-      this.tryNext(postRunInputs, { postRun: true })
+      Promise.resolve(false) :
+      this.maybeIncrementStep(postRunInputs, false)
     )
-    .then((stepRslt) => {
+    .then((didIncrementStep) => {
       // set `inProgress` to represent whether we are about to be inProgress
       // again (e.g. if pipeline will run again). in this regard cb()s get
       // accurate depiction of state
-      this.inProgress = !!stepRslt.didStep;
+      this.inProgress = !!didIncrementStep;
       this.events.emit('save-request', runOutput, null, forceSave);
       this.inProgress = false; // reset
-      if (stepRslt.didStep) {
+      if (didIncrementStep) {
         this.events.emit('inProgress', runOutput);
         return this.run(postRunInputs, compResult);
       }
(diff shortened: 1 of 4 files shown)


packages/coinstac-common/src/models/pipeline/pipeline.js
```diff
@@ -97,64 +97,81 @@ class Pipeline extends Base {
...
-          'pipelines do not permit concurrent running. please spawn a new',
-          'pipeline instance if concurrency is required',
-        ].join(' '))
-      );
+      return Promise.reject(new Error(
+        'Run inputs and ComputationResult required to pass to computations'
+      ));
+    } else if (this.inProgress) {
+      return Promise.reject(new Error(
+        /* eslint-disable max-len */
```
swashcap, line 119: When we update to the latest AirBnB styles we can drop the `max-len` disable. This is smart: makes it easier to grep/ag for strings in the codebase.

packages/coinstac-common/src/models/pipeline/pipeline.js
```diff
@@ -97,64 +97,81 @@ class Pipeline extends Base {
...
+      ));
     }
+
     this.inProgress = true;
     this.computation = this.computations[this.step];
-    let runCancelled = false;
     let forceSave = false;
-    // offer plugin hooks as functions.  restrict plugins from meddling with
-    // run state. simply expose some run modifiers.
+    let shouldContinueRun = true;
```
swashcap, line 128: Branching was all `!runCancelled`. Instead of negating the value on every `if`, it’s slightly less overhead to just flip the variable’s meaning.

I’m still a little iffy on the name, though. `shouldContinueRun` versus `shouldRunContinue` versus `isRunning`…

packages/coinstac-common/src/models/pipeline/pipeline.js
```diff
@@ -97,64 +97,81 @@ class Pipeline extends Base {
...
-    .then((runOutput) => this._postRun({
-      runInputs,
-      runOutput,
-      compResult,
-      runCancelled,
-      forceSave,
-    }))
-    .catch((err) => {
-      this.inProgress = false;
-      if (err) { err.isRunError = true; }
```
swashcap, line 155: Not sure what this was about. We’re always rejecting with an error or throwing in a `.then`. Might have been a vendor code thing?

packages/coinstac-common/src/models/pipeline/pipeline.js
```diff
@@ -212,40 +229,48 @@ class Pipeline extends Base {
...
+   * @param {boolean} [isPreRun=false] Whether method is called before
+   * `Pipeline#run` is called
+   * @returns {Promise<boolean>} Value representing whether the pipeline
+   * instance's `step` property was incremented and its `computation` property
+   * was updated: `true` indicates incrementation occurred, `false` indicates no
+   * change.
    */
-  tryNext(opts, cycle) {
-    const resp = { didStep: false };
+  maybeIncrementStep(opts, isPreRun) {
```
swashcap, line 266: Naming is _so_ instrumental in illuminating what these methods do. This method increments the instance’s `step` property and updates its `computation` property to match the step, but there’s some conditions to incrementation. It’s the _only_ method that effectively advances the pipeline’s current computation. Does `maybeIncrementStep` work? Could the name be clearer?

packages/coinstac-common/src/models/pipeline/pipeline.js
```diff
@@ -212,40 +229,48 @@ class Pipeline extends Base {
...
     const handleNextComplete = (doNext) => {
-      const endOfPipe = !this.computations[this.step + 1];
-      if (doNext && !endOfPipe) {
+      if (doNext && this.computations[this.step + 1]) {
         ++this.step;
         this.computation = this.computations[this.step];
-        resp.didStep = true;
+
+        return true;
```
swashcap, line 272: We were returning `{ didStep: true }` or `{ didStep: false }`, which is silly: it’s a boolean value. The extra abstraction of returning an object makes this method and its instance side-effects difficult to understand. If there‘s an in-between or third state we can switch to symbols.

packages/coinstac-common/src/models/pipeline/pipeline.js
rssk: Like `shouldRunCont` but nbd

packages/coinstac-common/src/models/pipeline/pipeline.js
rssk: I feel like I remember it being a vendor thing, forever ago

packages/coinstac-common/src/models/pipeline/pipeline.js
rssk: `conditionallyIncrement`? dunno this function is kinda convoluted, makes it harder to put meaning in an easy name

packages/coinstac-common/src/models/pipeline/pipeline.js
rssk: 👍

Issue #138, opened 2017-02-06
Add iteration count to computation results table


swashcap:
Add the computation’s iteration count to the results table in single-consortium view.

swashcap:
4d9cc59434a2e7cfbe0aaa27dc686ad59b24c783 took care of this.

PR #140, opened 2017-02-07
📝 Computation Output (take 2!)

swashcap:
* **Problem:**
  * Results reference the consortium document’s `activeComputationInputs` property, which is dynamic. This can cause interface explosions! :scream:
  * Computations need to be updated
  * Computation run meta is split across components
  * Result tables have a lot of data, making the interface cluttered
* **Solution:** Solve the problems!
* **Testing:** Run a couple computations. Confirm results look :+1:


swashcap:
Lil’ look at those collapsible results:

collapse-em

First one is un-collapsed by default.

swashcap:
4d9cc59434a2e7cfbe0aaa27dc686ad59b24c783 closes #138.

Issue #137, opened 2017-02-06
Drop computation run notifications


swashcap:
These fellas are a little chat-y:

![did-it-run-or](https://cloud.githubusercontent.com/assets/1858316/22667394/73066472-ec71-11e6-94b5-9070ed2d16bd.jpg)

everner-ds:
I recommend just having one box that updates with the computation step number. It could be a notification or be located somewhere else in the UI.

PR #141, opened 2017-02-08
Computation notifications → 🔕

swashcap:
* **Problem:** See #137.
* **Solution:** Drop comp run end notifications from _bg-services.js_
* **Testing:** Run a computation.


swashcap:
So much nicer now:

silence

:wave:

swashcap:
This PR modifies code minimally, so _I’m just doing it._ 💪

PR #144, opened 2017-02-13
Inter-process error notification

swashcap:
* **Problems:**
  * Main process errors don't always cause UI notifications
  * Projects with non-existent files :boom: explode in a non-useful way
* **Solutions:**
  * Add IPC event for async errors and wire up to UI error notification
  * `fs.stat` project’s files, throw a useful error when files DNE
* **Testing:**
  1. Spin up fresh client
  2. Create a project with a CSV from the _test_data_complete_ set.
  3. Try to run project
  4. Observe “Project files don't exist” error in UI

### DIFF
--- a/packages/coinstac-client-core/src/sub-api/project-service.js
+++ b/packages/coinstac-client-core/src/sub-api/project-service.js
@@ -91,6 +91,16 @@ class ProjectService extends ModelService {
           throw new Error('Expected covariates');
         }
 
+        return Promise.all([
+          project,
+          covariates,
+          ProjectService.statAll(
+            project.files.map(f => f.filename),
+            path.dirname(project.metaFilePath)
+          ),
+        ]);
+      })
+      .then(([project, covariates]) => {
         /**
          * [
          *   [1, 2, 3],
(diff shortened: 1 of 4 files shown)


packages/coinstac-ui/app/render/components/notification.js
```diff
@@ -18,6 +30,19 @@ export default class Notify extends React.Component {
...
+    ipcRenderer.removeListener('async-error', this.handleAsyncError);
+  }
+
+  handleAsyncError(event, message) {
+    this.push({
+      autoDismiss: 0,
+      level: 'error',
+      message: truncate(message, { length: 250 }),
+    });
+  }
```
cdaringe, line 44: maybe `console.error` it out too, but untruncated, that way if it actually _is_ suuuuper long, you can get all that great info you needed

swashcap: The full rejection/reasion is logged through some `app.logger` call (somewhere in main?) so we get the full message + stack trace.

packages/coinstac-client-core/src/sub-api/project-service.js
```diff
@@ -467,6 +477,33 @@ class ProjectService extends ModelService {
...
+   * @param {string} cwd Current working directory
+   * @returns {Promise<boolean>}
+   */
+  static statAll(files, cwd) {
+    const resolvedFiles = files.map(file => (
+      path.isAbsolute(file) ? file : path.resolve(cwd, file)
+    ));
+    const stat = bluebird.promisify(fs.stat);
+
+    return Promise.all(resolvedFiles.map((file) => stat(file).reflect()))
```
rssk, line 496: 👋

packages/coinstac-ui/app/render/components/notification.js
rssk: It is here: https://github.com/MRN-Code/coinstac/pull/144/files#diff-e68a4ebae3f693453666abc796d7a76cR17
:-)

PR #145, opened 2017-02-14
remove built tar first to prevent re-packing it

rssk:
Builds can grow in size if you are lazy and don't delete your tar'd build ----solution: rm it


swashcap:
:+1:

Issue #150, opened 2017-02-20
Alpha-sort FreeSurfer ROIs


swashcap:
Add a `.sort()` to the FreeSurfer valid ROIs in the computations:

https://github.com/MRN-Code/decentralized-laplacian-ridge-regression/blob/master/src/index.js#L33

https://github.com/MRN-Code/decentralized-single-shot-ridge-regression/blob/master/src/index.js#L139

PR #162, opened 2017-02-24
Update computations for ROI alpha-sorting

swashcap:
* **Problem:** See #150.
* **Solution:** Add FreeSurfer ROI alpha-sorting in computations. Bump computations' versions in COINSTAC.
* **Testing:** Check computations' ROIs in the UI.


swashcap:
Looks fine for both computations:

sorted-rois

Issue #157, opened 2017-02-20
Default computation ROI


swashcap:
Set the default ROI for single and multi-shot to `Right-Cerebellum-Cortex`. Add `todos` to remove this from the default. This probably should happen in the computations themselves.

PR #166, opened 2017-03-01
Bump computations' versions.

swashcap:
* **Problem:** See #157.
* **Solution:** Update computations' inputs, bump 'em
* **Testing:** Boot up UI and see ✨ _new defaults_ ✨


swashcap:
Lookin’ good:

lookin-good

Issue #151, opened 2017-02-20
Project computation inputs don’t change with consortium computation inputs


swashcap:
When a consortium owner adjusts computation inputs _or_ changes the computation entirely (single-shot → multi-shot) all users projects **do not** apply this change to their project data (specifically the `Project.metaCovariateMapping` property). This causes errors when a consortium owner runs a new computation and the users’ project’s don’t have the proper metadata CSV to computation input mappings.

ToDos:

- [x] Flag CSV mappings in the UI when the consortium changes
- [x] Don't allow user to contribute when mappings are bad

swashcap:
Thinking about solving this problem in a way that doesn't contribute to technical debt. The consortium document’s `activeComputationInputs` is the source of truth; when an owner modifies this property we need to ensure all projects associated with the consortium are marked invalid. Ideas:

* Store a complete copy of the consortium’s `activeComputationInputs` value on every user’s project document and deep compare the consortium’s and project’s values before contributing to a run. This is probably the easiest, but that’s quite a bit of data on the project document.
* Hash the consortium’s `activeComputationInputs` value by `JSON.stringify`-ing the data and place the hash on the project document. A client re-hashes the consortium document’s inputs value prior to run: if they hashes don’t match don’t allow the client to contribute to the run.
* Create consortia document change listeners on the clients that mark projects invalid when their associated consortium’s `activeComputationInputs` changes.

These strategies require pulling multiple documents to ensure a client _can_ contribute to a run. Application state becomes tied across multiple documents, unfortunately, which means multiple lookups and confusing code. Trying to determine another way…

swashcap:
We need a way to version _part_ of a document. We could attach a new integer `activeComputationInputsVersion` property to the consortium document, or place it inside the `activeComputationInputs` property somewhere. Still doesn’t get around the multiple doc lookups problem.

PR #168, opened 2017-03-02
Clear project metadata when computation input changes

swashcap:
### Problem

See #151.

### Solution

Add a consortium’s `activeComputationInputs` to a project. Don't allow a project to run when the consortium’s and project’s computation inputs don’t match. (See liberally commented code for details.)

#### Setup

1. Pull the branch:
    ```shell
    git fetch --all
    git checkout enhancement/#151-project-inputs
    ```
2. Install dependencies:
    ```shell
    npm run build
    ```
3. Boot COINSTAC’s UI
4. Create two consortia, ensuring each uses a different computation
5. Create a project

#### Test Cases

* A consortium’s computation inputs (ex: covariate names) change: the project should be flagged an un-runable
* An un-runable project’s CSV mapping `select`s should clear when the user edits the project
* A project’s selected consortium (from the dropdown) changes: the project’s CSV mapping `select`s should reset

### DIFF
--- a/packages/coinstac-client-core/package.json
+++ b/packages/coinstac-client-core/package.json
@@ -41,6 +41,7 @@
     "commander": "^2.9.0",
     "convict": "^1.3.0",
     "csv-parse": "^1.1.5",
+    "deep-equal": "^1.0.1",
     "halfpenny": "^6.0.0",
     "hawk": "^3.1.0",
     "inquirer": "^1.1.0",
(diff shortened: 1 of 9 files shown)


swashcap:
e4b29633694db33a0129373e07bade6efcc595a1 added this UI:

lil-note

The “Run” button is also disabled when covariates should be remapped.

PR #180, opened 2017-03-21
Fix UI consort create crash

### LINKED ISSUE #174: Un-initialized remote comp/consort state crashes the client UI
#problem
If there is a valid remote conn, but the DB has never be init'd by server-core, the add consort button crashes the UI

rssk:
With hardcoded comps that fail lookup, accessing a prop on that lookup
crashes the UI :-( 

Fix: make it null. This means the consortia page loads but you can't really do anything, better than crashing the whole app.

Closes #174


swashcap:
:+1: LGTM!

Issue #175, opened 2017-03-17
Comp results should ascending date order


rssk:
comp results should be the latest first

PR #182, opened 2017-03-23
Correct computation results' sort order.

swashcap:
* **Problem:** See #175.
* **Solution:**
  Sort computation results by their `endDate` property:

  a-few-wats-ago
* **Testing:** Run 2 computations in the UI.

### DIFF
--- a/packages/coinstac-ui/app/render/components/consortium/consortium-results.js
+++ b/packages/coinstac-ui/app/render/components/consortium/consortium-results.js
@@ -33,22 +33,26 @@ export default class ConsortiumResults extends Component {
     ) :
     (
       <ul className="list-unstyled">
-        {remoteResults.map((result, index) => {
-          const computation = computations.find(c => {
-            return c._id === result.computationId;
-          });
+        {remoteResults
+          .slice(0)
+          .sort(ConsortiumResults.compareRemoteResults)
+          .map((result, index) => {
+            const computation = computations.find(c => {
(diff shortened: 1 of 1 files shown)


packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -33,22 +33,26 @@ export default class ConsortiumResults extends Component {
     ) :
     (
       <ul className="list-unstyled">
-        {remoteResults.map((result, index) => {
-          const computation = computations.find(c => {
-            return c._id === result.computationId;
-          });
+        {remoteResults
+          .slice(0)
```
swashcap, line 37: Don’t mutate `remoteResults`!

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -68,3 +72,22 @@ ConsortiumResults.propTypes = {
...
 };
 
+/**
+ * Compare remote results for sorting purposes.
+ *
+ * @param {RemoteResult} a
+ * @param {RemoteResult} b
+ * @returns {number}
+ */
+ConsortiumResults.compareRemoteResults = (
```
swashcap, line 82: Should be easy to test this…when we get around to it.

swashcap:
Tested this on my machine. Lookin’ good!

Issue #181, opened 2017-03-21
Increase size on error log stack output


rssk:
Current error logs output limited traces, making them useless, increase ouput.

PR #183, opened 2017-03-23
Change log to output full error object.

swashcap:
* **Problem:** See #181.
* **Solution:** Use [serialize-error](https://www.npmjs.com/package/serialize-error) to include stack traces
* **Testing:** ?


swashcap:
Adding a `throw Error(/*...*/)` in _packages/coinstac-ui/app/render/log.js_ reveals this works:

hello-error

Issue #184, opened 2017-03-23
Switching computations disregards defaults in UI


swashcap:
Switching from single-shot to multi-shot in the consortium view causes multi-shot’s default computation input values to not be set in the UI. Note the empty “Lambda” field:

![no-lambda-gaa](https://cloud.githubusercontent.com/assets/1858316/24267083/4f5a6ea6-0fc6-11e7-910d-b16d1dda5c84.gif)

This causes problems when users re-map their covariates, and computation run attempts – via “Run Computation” button clicks – don’t appear to work correctly.

swashcap:
These run complications likely contributed to the extraneous dashboard items discovered on our 2017-03-22 run:

![so-status](https://cloud.githubusercontent.com/assets/1858316/24267204/d5303c54-0fc6-11e7-9807-ce1870ea1cd1.jpg)

PR #189, opened 2017-03-28
Bug/#184 computation switching

swashcap:
* **Problem:** The consortium form’s covariate state becomes irreparably contorted after computation change (see #184).
* **Solution:** Migrate all consortium form state to rely on redux-form. _(Bonus: keep consortium form state after user leaves form.)_
* **Testing:**
  1. Boot client
  2. Create a consortium with the _single-shot_ computation selected
  3. Save the consortium
  4. Edit the consortium to use the _multi-shot_ computation
  5. Run a computation


packages/coinstac-ui/app/render/components/computation-field-basic.js
```diff
@@ -64,13 +62,20 @@ export default class ComputationFieldBasic extends Component {
...
+       */
+      if (!Array.isArray(value) && value !== '') {
+        controlProps.value = [value];
+      }
+
       formControl = (
         <FormControl {...controlProps}>
           {options.map((option, index) => {
-            return <option key={index} value={index}>{option}</option>;
+            return <option key={index} value={option}>{option}</option>;
```
swashcap, line 78: No longer using indices as values and performing the following array lookup: just use the RoI value!

packages/coinstac-ui/app/render/components/computation-field-basic.js
```diff
@@ -7,54 +7,52 @@ import {
...
 export default class ComputationFieldBasic extends Component {
-  constructor(props) {
-    super(props);
-    this.handleButtonClick = this.handleButtonClick.bind(this);
-  }
-
   handleButtonClick(increment) {
-    const { max, min, onChange } = this.props;
-    const value = this.refs[INPUT_REF].props.value;
+    const {
```
swashcap, line 12: Method cleaned up to rely on Redux Form’s passed down properties instead of the homemade _refs_ approach.

packages/coinstac-ui/app/render/components/computation-field-covariates.js
```diff
@@ -1,144 +1,159 @@
...
-  handleNewClick() {
-    this.props.onChange(
-      {
-        name: '',
-        type: null,
-      },
-      this.props.value.length
-    );
-  }
+function covariateName({
```
swashcap: Couple additional components in this module which should be moved to their own files. We can tackle when we have time to refactor.

packages/coinstac-ui/app/render/components/consortium/consortium-form.js
```diff
@@ -76,102 +75,52 @@ class ConsortiumForm extends Component {
...
+    const covariatesIndex = values.activeComputationInputs.findIndex(input => (
+      Array.isArray(input) &&
+      input.length &&
+      typeof input[0] === 'object'
+    ));
+
+    if (
+      covariatesIndex > -1 &&
+      values.activeComputationInputs[covariatesIndex].some((covariate) => !(
+        covariate.name.length &&
```
swashcap: These checks become a little gross, but Redux Form doesn't seem to like passing configuration with props. `values` contains only form input.

swashcap:
Rebased this on `master` and now I see this when attempting to add a metadata file:

wat-is-get

`master` itself is good. Might have something to do with how covariates are saved?

swashcap:
Ah, shucks. Looks like a project component was relying on `ComputationFieldCovariates.typeMap`, which this PR dropped. Re-adding it.

swashcap:
Looking good now:

work-it

Issue #186, opened 2017-03-23
Incorrect result betas order 


swashcap:
Investigate the result betas’ order in the computation result table:

![beta-order](https://cloud.githubusercontent.com/assets/1858316/24267402/6286fdf4-0fc7-11e7-8c7a-526f1b48c58b.jpg)

The “isControl” and “Age” labels should be reversed?

Issue #12, opened 2016-07-12
Computation Inputs and Outputs


swashcap:
## Problem

Computations need a codified method of requesting information from clients and outputting the information of a run. Specifically, the ridge regression algorithm needs a collection of files from each client. Files must be tagged “control” or “not control.”

The need for input from clients is matched with a related requirement on the remote end: the need for output. While nothing in the ecosystem exists that requires an output, consider the ridge regression has certain key values (r², gradient, objective) that are deemed important. Also, consider an algorithm that maps computation to a 2D scatterplot – like Sergey’s presentation from a couple weeks ago. This could eventually land in COINSTAC.
## Goals
1. Provide a straightforward way for algorithm authors to declare inputs (files). These inputs must be qualifiable by the clients (tagging files).
2. Create dialogs and interfaces in _coinstac-ui_ that support algorithm inputs.
3. Consider future outputs, such as display of numbers and text or visualizations. While this isn’t currently necessary, we should strive for harmony between input and output syntax/methodology.
4. Make it extensible: algorithm authors should have maximum flexibility when declaring inputs. What types of files does an algorithm require? Should we support more than boolean tagging? (`isControl: true|false` versus `isControl: 'my-string'`.) Will files need to be grouped together? Can we leverage the way researchers store files in directories?
## Potential Solution

Create `input` and `output` keys in computation definitions, and design new schemas for each.

Here’s a sample computation definition:

``` js
module.exports = {
  name: 'my-computation',
  version: '1.0.0',
  cwd: __dirname,
  local: [{
    type: 'function',
    verbose: false,
    fn: opts => {
      return Math.random();
    },
  }],
  remote: [{
    type: 'function',
    verbose: true,
    fn: opts => {
      const target = 0.618;

      if (opts.userResults.some(n => Math.round(n * 1e3) / 1e3 === target)) {
        return { complete: true };
      }
    },
  }],
};
```

(See [decentralized-laplacian-ridge-regression](https://github.com/MRN-Code/decentralized-laplacian-ridge-regression/blob/master/src/index.js) for a full example.)

We can extend the `local` and `remote` computation objects like so:

``` js
{
  // ...
  local: [{
    type: 'function',
    fn: // ...
    input: {
      type: 'files',
      tags: ['isControl'],
    },
  }],
  // ...
}
```

``` js
{
  // ...
  remote: [{
    type: 'function',
    fn: // ...
    output: {
      type: 'function',
      fn: opts => {
        return {
[...]

PR #203, opened 2017-04-10
Correct covariates’ order in results table

swashcap:
* **Problem:** covariates’ order in the results table is incorrect (see #186).
* **Solution:** Sort covariates in the UI to match their sort order in the computation:
  1. The client [converts covariates to camelCase](https://github.com/MRN-Code/coinstac/blob/39f4035e00558746ba9687c36024ed46a254f73c/packages/coinstac-client-core/src/sub-api/project-service.js#L180) in each file’s `.tags` map
  2. The computations [alpha-sort the each file’s `.tags`’s properties](https://github.com/MRN-Code/decentralized-single-shot-ridge-regression/blob/d4a5b270a085ce191f1a24cae729fd89500fa65a/src/index.js#L49)
* **Testing:**
  1. Run a computation
  2. Ensure results table’s covariates are alpha-sorted

### DIFF
--- a/packages/coinstac-ui/app/render/components/consortium/consortium-result.js
+++ b/packages/coinstac-ui/app/render/components/consortium/consortium-result.js
@@ -2,7 +2,7 @@ import React, { PropTypes } from 'react';
 import { Collapse, Label } from 'react-bootstrap';
 import classNames from 'classnames';
 import moment from 'moment';
-import { reduce } from 'lodash';
+import { camelCase, reduce } from 'lodash';
 
 import ConsortiumResultMeta from './consortium-result-meta';
 import ConsortiumResultTable from './consortium-result-table';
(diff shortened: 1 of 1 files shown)


swashcap:
This solution works with “A Control” and “Z Age”:

alpha-sortz

swashcap:
We want to abstract out FreeSurfer parsing and user input mapping (see #12) in the near future, ensuring computations deal with the raw numbers and labels (betas/covariates/whatever) are properly re-mapped for the output. This as a temporary fix until we have time to focus on that.

packages/coinstac-ui/app/render/components/consortium/consortium-result.js
```diff
@@ -61,10 +61,15 @@ export default function ConsortiumResult({
...
-    const covariates =
+    const covariatesIndex =
       computation.name === 'decentralized-single-shot-ridge-regression' ?
-      computationInputs[0][2].map(x => x.name) :
-      computationInputs[0][3].map(x => x.name);
+      2 :
+      3;
+
+    const covariates = computationInputs[0][covariatesIndex]
+      .slice(0)
```
swashcap, line 70: Don’t mutate!

JingMing623:
Now, the order of the output beta vector are correct.

swashcap:
I confirmed this works yesterday. Mergin’ 💖

Issue #192, opened 2017-03-28
Clarify “Global” computation output


swashcap:
The computation output tables’ “Global” label wasn’t clear to Jess in our 2017-03-27 demo. She asked whether this was “meta analysis” or just an “average.” How can we clarify the output’s methods to users?

swashcap:
@JingMing623 @everner-ds Any thoughts on this? Should we ignore for the demo as we can explain output on a call?

JingMing623:
For single-shot and multi-shot, the word 'global' indeed has different meanings.  for single-shot will be meta-analysis result (average), and for multi-shot, it is mega-analysis. It is kind of depending on which algorithm we are using.  If we want to change 'Global' to other words, probably need to link to different algorithm too.

swashcap:
We could temporarily add a tooltip to the “Global” headings to clarify its section’s meaning:

![meta-analysis-whoa](https://cloud.githubusercontent.com/assets/1858316/24923961/0cc6dbe2-1ea8-11e7-8e87-2248541d57fa.gif)

In the future we should provide a way for computation authors to explain their computations’ output in a sentence or two. Perhaps this can go in the computation’s root `module.exports`:

```js
// computation.js

module.exports = {
  name: 'Multi-Shot Ridge Regression',
  version: '1.0.2',
  descriptions: {
    main: 'Iterative decentralized ridge regression.',
    output: 'Mega-analysis',
  },
};
```

This could also go in a _package.json_ entry point.

JingMing623:
Looks nice! Yes, I agree computation author should provide this information to coinstac

swashcap:
Closing this issue. We can revisit advanced description output in #12.

PR #205, opened 2017-04-11
Add tooltip to results’ “Global” heading

swashcap:
* **Problem:** The “Global” heading in the computation result table doesn’t make sense to users (see #192).
* **Solution:** Clarify the headings’ meanings with a computation-aware tooltip
* **Testing:**
  1. Run a computation in the UI
  2. Observe the tooltip in the result table


swashcap:
Here’s what it looks like:

Tooltip UI

PR #207, opened 2017-04-18
Add overview document

swashcap:
Opening a preliminary PR for an “overview”-type, informal system doc. How’s it looking?


rssk:
I think this looks good, short and to the point

swashcap:
Going ahead with the merge. We can revise in future PRs.

PR #209, opened 2017-04-27
Bugfix/consortium form clearing

swashcap:
#### Problem

Changing active computations in the consortium form clears the label and description.

where-it-go

#### Solution

Use [redux-form’s `formValueSelector`](http://redux-form.com/6.6.3/docs/api/FormValueSelector.md/) to persist dirty form state in the `initialValues` property.

#### Testing

1. Boot the UI
2. Open the consortium creation form
3. Fill out “label” and “description” fields
4. Change the active computation
5. Observe “label” and “description” aren’t cleared!


swashcap:
Confirmed this works and saving persists the documents:

yep-looking-good

Going ahead with the merge.

PR #220, opened 2017-05-22
Enhancement/issues 218 219

tkah:
*  **Problems:**
    * Status-item button easily overlooked
    * Status-item button takes user to top of consortium page, rather than results section as expected
* **Solution:**
   * In addition to existing link, Bootstrap button added to bottom-right of status-item
   * Status-item link includes computation id to automatically expand the selected computation
   * ID added to results section to bring user directly to results section
* **Tests:**
   1. Login into application and note placement of new buttons in completed status items listed on home screen
   1. Click one of the newly added buttons and check that not only does the next page bring you to the **Results** section, but, once the data resolves, the selected computation is automatically expanded.


swashcap:
@tkah For future reference, please address multiple issues with multiple pull requests. Targeting a PR to a specific issue makes review easier, cleaner history, and reversion effortless.

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -1,33 +1,43 @@
...
     }
+  }
 
-    this.setState({
-      expanded: this.state.expanded,
-    });
+  toggleCollapse(computationId) {
+    const { setExpandedComputation, expandedComputation } = this.props;
+
+    if (computationId !== expandedComputation) setExpandedComputation(computationId);
```
swashcap: Surprised the linter’s not barking at this, although the style guide doesn’t appear to have strong opinion around [blocks](https://github.com/airbnb/javascript#blocks). Please add explicit braces around `if`s to keep things readable and consistent:

```diff
- if (computationId !== expandedComputation) setExpandedComputation(computationId);
- else setExpandedComputation('');
+ if (computationId !== expandedComputation) {
+   setExpandedComputation(computationId);
+ } else {
+   setExpandedComputation('');
+ }
```

Or, use a ternary operator:

```js
setExpandedComputation(
  computationId !== expandedComputation ? computationId : ''
);
```

packages/coinstac-ui/app/render/state/ducks/consortia-page.js
```diff
@@ -0,0 +1,21 @@
+const SET_EXPANDED_COMPUTATION = 'SET_EXPANDED_COMPUTATION';
```
swashcap: This is fine for now, but doesn’t account for multiple expanded computations across multiple consortia pages. We likely want to track this state at some point so users return to UI in a familiar condition.

packages/coinstac-ui/app/render/styles/app.scss
```diff
@@ -142,7 +142,7 @@ $icon-font-path: "~bootstrap-sass/assets/fonts/bootstrap/";
     width: 1em;
     height: 1em;
     margin: 10px 10px;
-    position: absolute;
+    position: fixed;
```
swashcap, line 145: 👀

swashcap (changes requested):

packages/coinstac-ui/app/render/routes.js
```diff
@@ -27,7 +27,7 @@ export default (
       <IndexRoute component={DashboardHome} />
       <Route path="/consortia" component={DashboardConsortia} />
       <Route path="/consortia/new" component={ConsortiumController} />
-      <Route path="/consortia/:consortiumId" component={ConsortiumController} />
+      <Route path="/consortia/:consortiumId(/:computationId)" component={ConsortiumController} />
```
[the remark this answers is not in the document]
tkah: When I wrote this, I was just using the optional route param _computationId_ (now _resultId_) to autoexpand the respective result accordion and not to jump to that section of the page, though the capacity was there. I've now built it to work that way using a timeout, which feels hacky, but I've retained the _resultId_ as a route param. This might not follow the web paradigm for anchoring to a section of a "view", but since that's not exclusively its purpose, is that okay?

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -19,14 +19,15 @@ class ConsortiumResults extends Component {
...
-    }
+    const { expandedResult } = this.props;
+
+    // Hacky - used to delay querying ids until results list items are rendered
+    setTimeout(() => {
+      if (document.querySelector(`#C${expandedResult}`)) {
+        document.querySelector(`#C${expandedResult}`).scrollIntoView();
+      }
+    }, 500);
   }
```
tkah: I used setTimeout here to wait to query the result id until after the results list items were rendered. I'm sure there is a better way and would welcome some advice.

Also, the prepended 'C' in the ID is to avoid an error being thrown when using _querySelector_ on an ID that can begin with a digit. I learned today that is a no, no.

packages/coinstac-ui/app/render/routes.js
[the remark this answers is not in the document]
swashcap: Yeah, I think it’s okay for now.

swashcap:
Noticing the expand/collapse doesn’t target a specific result:

double-thingy

Both expand/collapse when only one trigger is clicked.

packages/coinstac-ui/app/render/components/consortium/consortium-controller.js
```diff
@@ -181,11 +184,11 @@ function mapStateToProps(state, { params: { consortiumId } }) {
     computations: (computations || [])
       .sort((a, b) => `${a.name}@${a.version}` > `${b.name}@${b.version}`),
     consortium,
+    initialResultId,
```
swashcap: You can write:

```diff
- initialResultId
+ initialResultId: resultId
```

here. No need to create a new variable for it.

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -1,33 +1,47 @@
...
+  componentWillMount() {
+    const { initialResultId, setExpandedResult } = this.props;
+
+    setExpandedResult(initialResultId);
+  }
+
+  goToResults() {
+    const { expandedResult } = this.props;
+
+    // Hacky - used to delay querying ids until results list items are rendered
```
swashcap: I agree: not ideal, but works for now. Could you change this comment to include the word `TODO:`? That way, when we search the codebase we can easily identify this.

Good example: https://github.com/MRN-Code/coinstac/blob/eb1c71bb394e091e811a1cfa3f665c559aa31d65/packages/coinstac-ui/app/render/components/settings.js#L24-L25

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -92,3 +110,11 @@ ConsortiumResults.compareRemoteResults = (
 
   return 0;
 };
+
+const mapStateToProps = ({ consortiaPage }) => {
```
swashcap: You can destructure this directly in the parameter:

```js
const mapStateToProps = ({
  consortiaPage: { expandedResult },
}) => {
  /* ... */
};
```

Avoids the variable declaration line.

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -2,46 +2,52 @@ import React, { Component } from 'react';
...
+import { setExpandedResults } from '../../state/ducks/consortia-page';
 
 import ConsortiumResult from './consortium-result';
 
 class ConsortiumResults extends Component {
   constructor(props) {
     super(props);
 
+    this.state = { wentToResult: false };
     this.goToResults = this.goToResults.bind(this);
```
tkah: *wentToResult* is a flag on the current state to prevent the page from jumping each time a new render cycle is encountered. Just indicates that the *goToResult()* function has already been called.

tkah:
I'm having a hard time recreating the expand/collapse behavior. Still looking into it. Still, I realized that only one accordion would be expandable at a time with the previous version. This should now allow multiple accordions to be expanded at once.

packages/coinstac-ui/app/render/state/ducks/consortia-page.js
```diff
@@ -1,20 +1,29 @@
-const SET_EXPANDED_RESULT = 'SET_EXPANDED_RESULT';
+const SET_EXPANDED_RESULTS = 'SET_EXPANDED_RESULT';
```
swashcap: It's nice for the variable name to match the action name. Makes debugging easier.

swashcap:
Linting and tests work for me. Not quite sure what’s upsetting Codeship. :+1:

swashcap:
Sneaking b76e63f81ae7a8e04ed8eca851160e08d82f4c7b in there. ¯\\\_(NPM 5)_/¯

swashcap:
Clicking expand/contract still effects both result items:

both-contract

swashcap (approved):

packages/coinstac-ui/app/render/components/consortium/consortium-results.js
```diff
@@ -61,11 +61,11 @@ class ConsortiumResults extends Component {
             });
 
             return (
-              <li key={index} id={`C${result.computationId}`}>
+              <li key={index} id={`C${result._id}`}>
```
swashcap: :+1:

swashcap:
Seeing this when I click the “Add Consortium” button:

warn-warn

@tkah Can you make sure the empty prop isn't passed down? These errors because useless noise when debugging.

swashcap:
Otherwise, it’s working great:

itsworking

🎉

swashcap:
Sorry for the delay on this. Looks great, nice work! 🍾

PR #221, opened 2017-05-23
Whitelist new `user` model properties.

swashcap:
* **Problem:** New `user` model properties added in MRN-Code/quarterback#127 et. al. and in the future MRN-Code/quarterback#58 break COINSTAC login requests due to strict model property validation.
* **Solution:** Whitelist new properties
* **Testing:**
  * `npm run lint && npm test` on this branch
  * Log in on an API server with _and_ without the new `user` model properties


swashcap:
Looks like Codeship’s python env’s all screwy: https://gist.github.com/swashcap/13f8f839c69340c48352ea821dc1dad6. _Gross._

PR #222, opened 2017-05-26
moved consortia results call from dashboard-home to redux action

tkah:
* **Problem:**
   1. Dashboard status-items are loaded slowly after initial load despite the fact that they should be quickly retrievable from local storage.
   1. Dashboard displays *No computation results* message and no spinner despite the fact that there are computation results still loading.
* **Solution:**
   1. Changed remote results request to pull from redux action and add to props rather than pulling on each load and adding results to state.
* **Tests:**
   1. Login to application and note the status-items loading on the right-side panel.
   1. Navigate into the *Consortia* tab, then back into the *Home* tab. Results should load immediately if model still exists.
   1. Navigate into the any of the status-items' results, then navigate back to *Home*. If model reloads, you should see the spinner rotating until remote results have been retrieved.


tkah:
It's possible that this PR introduces problems elsewhere and I'd appreciate any clarification.

packages/coinstac-ui/app/render/components/dashboard-home.js
```diff
@@ -2,76 +2,32 @@ import React, { Component } from 'react';
...
         logger.error(error);
         notify('error', error.message);
       }
     }));
-
-    this.getRemoteResults().then(() => {
-      this.interval = setInterval(() => this.getRemoteResults(), 2000);
-    });
-  }
-
```
tkah, line 36: Much of this section of code deals with setting and clearing out the *remoteResults* interval. Since *remoteResults* are now being pulled from redux, the timer could just call that action. I just wanted to double-check that there isn't some back-end process that could update this before I put the interval timer back in.

@rssk indicated I might try something akin to the syncEmitters in db-services. After poking around a little, setting an emitter on a dynamically named db might be problematic? Would https://pouchdb.com/guides/changes.html work any better?

packages/coinstac-ui/app/render/components/dashboard-home.js
```diff
@@ -2,76 +2,32 @@ import React, { Component } from 'react';
...
   componentWillMount() {
-    const { dispatch } = this.props;
+    const { computations, dispatch, remoteResults } = this.props;
+
+    if (!remoteResults.length) {
+      dispatch(fetchConsortiaResults());
+    }
+
```
tkah: Using a redux action means that the spinner will appear while the application waits for *remoteResults* to resolve.

packages/coinstac-ui/app/render/state/ducks/remote-results.js
```diff
@@ -20,6 +20,24 @@ export const fetch = applyAsyncLoading(id => (dispatch) => {
...
+      ))
+    )
+    .then((responses) => {
+      const results = sortBy(flatten(responses), ['endDate', 'startDate']).reverse();
+      dispatch(setRemoteResults(results));
+      return results;
+    });
+  };
+});
+
```
tkah: Perhaps there needs to be a separate action for results from all consortia versus results from a single consortium?

packages/coinstac-ui/app/render/components/dashboard-home.js
```diff
@@ -2,76 +2,32 @@ import React, { Component } from 'react';
...
-    const { dispatch } = this.props;
+    const { computations, dispatch, remoteResults } = this.props;
+
+    if (!remoteResults.length) {
+      dispatch(fetchResults());
+    }
+
+    if (!computations.length) {
+      dispatch(fetchComputations());
+    }
```
tkah, line 30: Moving results fetching to redux allows the spinner to keep spinning while *remoteResults* is being grabbed.

tkah:
This may also negate the need for the timeout in PR #220.

packages/coinstac-ui/app/render/components/dashboard-home.js
rssk: It would be you'd have to know what to filter by

swashcap:
Finally getting around to testing this. The top right loading spinner firing every ~3 seconds is _super annoying:_ this is a global indicator that fires when view change (ex: “My Files” to “Consortia”) or when a view fetches all its data (single consortium loads). It sends the wrong signal to the user when it fires so frequently: we’re simply updating the count. Ideally we provide a local indicator that signifies “this is loading” in the dashboard card.

Also, that dang loader still displays after a computation’s completely done:

still-going

_Annoying._ Let’s see what we can do to make this better.

swashcap:
It’s loading in the empty state:

empty-rando-loadin

😳 Also, seeing this error as I navigate around in a running client:

```
warning.js:36 Warning: Failed prop type: The prop `data.global` is marked as required in `ConsortiumResult`, but its value is `undefined`.
    in ConsortiumResult (created by ConsortiumResults)
    in ConsortiumResults (created by Connect(ConsortiumResults))
    in Connect(ConsortiumResults) (created by Consortium)
    in div (created by Consortium)
    in Consortium (created by ConsortiumController)
    in ConsortiumController (created by Connect(ConsortiumController))
    in Connect(ConsortiumController) (created by RouterContext)
    in div (created by Dashboard)
    in div (created by Dashboard)
    in div (created by Dashboard)
    in Dashboard (created by Connect(Dashboard))
    in Connect(Dashboard) (created by RouterContext)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by RouterContext)
    in RouterContext (created by Router)
    in Router (created by Root)
    in Provider (created by Root)
    in Root
```

Circumstances:

bugzzzz

swashcap:
Okay, looking into this. So far, so good. Did see this on a client during multi-shot:

```
"Error: Cannot get property 'name' on missing remote object 833
    at throwRPCError (/Users/creed/coinstac/packages/coinstac-ui/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/browser/rpc-server.js:145:17)
    at EventEmitter.<anonymous> (/Users/creed/coinstac/packages/coinstac-ui/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/browser/rpc-server.js:391:7)
    at WebContents.<anonymous> (/Users/creed/coinstac/packages/coinstac-ui/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/browser/api/web-contents.js:232:13)"
```

I think this is an Electron bug.

swashcap:
Single-shot and multi-shot function fine. Nice work!

PR #226, opened 2017-06-12
Document Conflict In Multi-Site Run

### LINKED ISSUE #224: Doc conflicts for owners on run
When an owner starts a ridge-regression run, they receive notifications for a `406 Document conflict` from pouch. However the run _seems_ to continue fine.

tkah:
* **Problem:** Running a ridge regression with multiple sites can result in those sites receiving a *Document Conflict* error.
* **Solution:** Wrap `joinSlavedRun` in `else` statement.
* **Test:** Run a ridge regression with multiple clients. Run a second time after the first completes and notice no *Document Conflict* errors.

Resolves #224


swashcap:
Geez, GitHub’s diff is confusing. Tested this via the UI with 2 clients. It works! 🎉

rssk:
The diff is bad... This is my fault, a run operation was being called twice, resulting in two updates

PR #227, opened 2017-06-13
remove check from lower level common, keep mem/disk check in config

rssk:
#Problem
Hey, state clearing isnt working....

#Solution
It appears that the check for which mem/disk adapters we use for the DB's happens in two places (kinda). Simplify that logic, make it one place more or less, and blamo, fixed.


swashcap:
Ran a computation via the UI. Only on-disk databases are the ones we want:

```shell
$ ls -1 ~/.coinstac/demo-1/dbs
local-consortium-c8e4797c948441d4bc91425b4477a5cb/
projects/
```

:tada:

PR #233, opened 2017-06-19
Enhancement/end to end testing

tkah:
* **Problem:**
   1. Need end-to-end user tests.
* **Solution:**
   1. Write 'em using Spectron, Chai and Mocha
* **Tests:**
   1. Install new dependencies: `cd packages/coinstac-ui`, `npm install`
   1. From coinstac-ui directory: `npm run e2e`


packages/coinstac-ui/package.json
```diff
@@ -91,7 +95,8 @@
     "clean:db": "node scripts/clean-db.js",
     "docs": "mkdir -p docs && jsdoc -t ./node_modules/minami -d docs -R README.md -r src/",
     "test": "nyc --reporter=lcov --check-coverage --lines=1 --functions=1 --branches=1 babel-node test/",
-    "watch": "webpack-dashboard -p 3001 -- node webpack-dev-server.js"
+    "watch": "webpack-dashboard -p 3001 -- node webpack-dev-server.js",
+    "e2e": "mocha --timeout=120000 test/e2e/test.js"
```
swashcap: It’s often nice to namespace testing scripts with `test`, so this should be `test:e2e`. See [coins-pr-site’s scripts](https://github.com/MRN-Code/coins-pr-site/blob/dacdab3de4c02d4f1ec402fef15c8e83f5697f19/package.json#L102-L108) for a good example. I also like to alphabetize the `scripts` properties to match how npm automatically alphabetizes `dependencies` and `devDependencies`.

packages/coinstac-ui/test/e2e/test.js
```diff
@@ -0,0 +1,143 @@
...
+const mocksPath = path.join(__dirname, 'mocks.js');
+
+const EXIST_TIMEOUT = 4000;
+const USER_ID = 'test';
+const PASS = 'test';
+const CONS_NAME = 'e2e-consortium';
+const CONS_DESC = 'e2e-description';
+const PROJECT_NAME = 'e2e-files';
+
+global.before(function () {
```
swashcap: You should be able to drop the `global` here: the engine should resolve `before` to Mocha's global function. Specifying mocha as an [ESLint env config](http://eslint.org/docs/user-guide/configuring#specifying-environments) might help if you’re seeing lint errors.

Also, because `chai.should` and `chai.use` are synchronous you shouldn’t need to wrap them in a `before`.

packages/coinstac-ui/test/e2e/test.js
```diff
@@ -0,0 +1,143 @@
...
+  });
+
+  it('displays the correct title', function () {
+    return app.client.waitUntilWindowLoaded()
+      .getTitle().should.eventually.equal('COINSTAC');
+  });
+
+  it('authenticates demo user', function () {
+    return app.client
+            .setValue('#login-username', USER_ID)
```
swashcap: :eyes: _Indentation?_

packages/coinstac-ui/test/e2e/test.js
```diff
@@ -0,0 +1,143 @@
...
+describe('Testing::e2e', function () {
+  before(function () {
+    return app.start();
+  });
+
+  after(function (done) {
+    app.stop().then(function () { done(); });
+  });
+
+  it('opens a single window', function () {
```
swashcap: Why do the `it` blocks need their own function scope?

swashcap (changes requested):

packages/coinstac-ui/package.json
swashcap: Also, end-to-end testing relies on built front-end assets: an empty _app/render/build/_ results in test failure with no appropriate error message. The end-to-end script should include asset builds to eliminate this case.

swashcap:
Few minor details to fix, but otherwise it works great!

its-working

🎉

packages/coinstac-ui/test/e2e/test.js
```diff
@@ -0,0 +1,140 @@
...
+chai.use(chaiAsPromised);
+
+const app = new Application({
+  path: electronPath,
+  env: { NODE_ENV: 'test' },
+  args: [appPath, '-r', mocksPath],
+});
+
+describe('Testing::e2e', () => {
+  before(() => {
```
swashcap: You can drop the `return` and use the arrow function’s implicit return:

```diff
-  before(() => {
-    return app.start();
-  });
+ before(() => app.start());
```

Not sure why the linter isn’t prompting you to do this (probably need to upgrade).

packages/coinstac-ui/test/e2e/test.js
```diff
@@ -0,0 +1,140 @@
...
+    return app.start();
+  });
+
+  after((done) => {
+    app.stop().then(() => { done(); });
```
swashcap: [`spectron.Application#stop` returns a `Promise`](https://github.com/electron/spectron#stop), so you don’t need to use a callback here. From [spectron’s example usage](https://github.com/electron/spectron#using) they check `spectron.Application#isRunning`, too. You can write this as:

```js
after(() => {
  if (app && app.isRunning()) {
    return app.stop();
  }
});
```

swashcap (changes requested):

swashcap (approved):

swashcap:
```shell
$ time npm run test:e2e
# ...
# ...
# ...
real    6m2.793s
user    3m44.403s
sys     0m34.630s
```

😬 We can look into speeding this up later.

PR #236, opened 2017-07-10
Refactor/consistent action dispatch

tkah:
* **Problem:**
   1. Redux actions are not dispatched consistently. 
   1. Component database calls would be better served by already existing actions.
* **Solution:**
   1. Use shorthand pre-bound action creators as outlined [here](http://blog.isquaredsoftware.com/2016/10/idiomatic-redux-why-use-action-creators/).
   1. Call actions over separate database calls. (Really only one in `coinstac-ui/app/render/components/projects/form-project-controller.js`)
* **Tests:**


packages/coinstac-ui/app/render/components/user-account-controller.js
```diff
@@ -38,7 +37,7 @@ class UserAccountController extends Component {
 
   render() {
     return (
-      <UserAccount logout={this.logout} {...this.props} />
```
swashcap: Why change this property’s name?

packages/coinstac-ui/app/render/components/dashboard-consortia.js
```diff
@@ -23,12 +23,12 @@ class DashboardConsortia extends Component {
 
           return (
             <ConsortiumCard
-              deleteConsortium={() => dispatch(deleteConsortium(id))}
```
swashcap: In cases where you’re using multiple properties/functions from `this.props` it’s nice to destructure:

```js
const {
  deleteConsortium,
  joinConsortium,
  leaveConsortium,
} = this.props;

// ...
 <ConsortiumCard
  deleteConsortium={() => deleteConsortium(id)}
  joinConsortium={() => joinConsortium(id, username)}
  leaveConsortium={() => leaveConsortium(id, username)}
/>
```

A little nicer to read.

packages/coinstac-ui/app/render/components/consortium/consortium-controller.js
```diff
@@ -151,15 +151,23 @@ ConsortiumController.contextTypes = {
 ConsortiumController.displayName = 'ConsortiumController';
 
 ConsortiumController.propTypes = {
+  addConsortiumComputationListener: PropTypes.func,
```
swashcap: Holy additional prop types. 😲

swashcap (changes requested):

packages/coinstac-ui/app/render/components/consortium/consortium-controller.js
tkah: Yeah, using this manner of binding means more prop types unfortunately.

packages/coinstac-ui/app/render/components/dashboard-consortia.js
tkah: 👍

packages/coinstac-ui/app/render/components/user-account-controller.js
tkah: Adding a new prop called `logout` to take the place of the previously dispatched action `logout` meant that when we pass into `UserAccount` a function named `logout` referencing AccountController's local function in addition to using the spread operator to pass in all the props, including the new `logout` prop, we see issues. The UserAccount component would have two props named `logout`. So, I just renamed the local function.

swashcap:
@tkah Want to resolve the conflicts? Then I can test and merge.

packages/coinstac-ui/app/render/components/user-account-controller.js
swashcap: Looks like `UserAccount` only needs a `user` prop, but that makes total sense. We might want to pass more stuff from `UserAccount Controller` to `UserAccount` in the future.

packages/coinstac-ui/app/render/components/consortium/consortium-controller.js
swashcap: Comes with the territory I guess :+1:

swashcap (approved):

packages/coinstac-ui/app/render/components/dashboard-home.jsx
```diff
@@ -21,33 +20,38 @@ class DashboardHome extends Component {
...
   componentWillUpdate() {
     const {
       computations,
       consortia,
-      dispatch,
+      fetchRemoteResultsForUser,
       username,
     } = this.props;
 
-    console.log('update');
```
swashcap, line 41: Whoa, nice catch.

PR #237, opened 2017-07-10
changed test to allow for existing consortia

tkah:
* **Problem:**
   1. Tests fail when testing against system with already existing consortia.
* **Solution:**
    1. Specify which consortium is to be deleted/created.
* **Tests:**
   1. Create separate consortium, project


packages/coinstac-ui/app/render/components/consortium-card.js
```diff
@@ -25,7 +25,7 @@ export default function ConsortiumCard(props) {
 
   if (isOwner) {
     deleteButton = (
-      <Button bsSize="small" bsStyle="danger" onClick={deleteConsortium}>
+      <Button bsSize="small" bsStyle="danger" onClick={deleteConsortium} id={`delete-${label}`}>
```
swashcap: A consortium’s `label` may have whitespace, which means selection by ID could break. Perhaps change to the consortium’s `id`?

swashcap (changes requested):

packages/coinstac-ui/app/render/components/consortium-card.js
tkah: The reason I've added these `id` tags is to be able to grab them from WebdriverIO. In the e2e test, I don't have access to the `id` of a given consortium. I have the name I assign to it, though. Would you be alright with using a modified `label` with spaces replaced by dashes?

swashcap: Ah, gotcha. That sounds like a good approach.

swashcap (approved):

swashcap:
@tkah This looks good. Could you fix the merge conflicts?

swashcap:
🎉 Works great!

PR #238, opened 2017-07-11
📄 Architecture requirements

swashcap:
* **Problem:** The repository doesn’t have an architecture requirements document!
* **Solution:** Add it from the internal Google doc. Do some cleanup.
* **Testing:** n/a

### DIFF
--- a/design/architecture-requirements.md
+++ b/design/architecture-requirements.md
@@ -144,8 +144,7 @@ A client is the local application that users interact with to use COINSTAC.
 11. Should afford users a reasonable measure of security
     1. User information, consortium information, PHI data should be handled correctly
     2. E.g., Someone creates a computation that sends everyone’s data to another server
-12. Error handling
-    1. Sometimes, a crash in one step might be related to errors in a previous step. The user should be able to manually send free-form text feedback and automated crash reports to the author of any computation.
+12. Error handling: a crash in one step might be related to errors in a previous step. The user should be able to manually send free-form text feedback and automated crash reports to the author of any computation.
 
 ### System Characteristics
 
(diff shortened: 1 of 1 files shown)


design/architecture-requirements.md
```diff
@@ -0,0 +1,169 @@
...
+    1. Debugging tools (Example: coinstac-simulator) should exist to allow authors to easily test their algorithms and simulate their use within a COINSTAC-like system
+    2. COINSTAC should perform automated checks against authors’ code to ensure it works within the system, and alert them to errors or potential problems.
+5. Should have an input and output specifications (see [MRN-Code/coinstac#12](https://github.com/MRN-Code/coinstac/issues/12)) to aid in composability and reproducibility:
+    1. Should be able to specify inputs: the computation dictates what it needs to run. This could be a collection of FreeSurfer files, a range of DICOM files, a CSV full of metadata, etc. This may also include inputs from previous computations in the pipeline (Example: second computation needs the original files for meta-analysis).
+    2. Should be able to specify outputs: the computation declares its results. In some cases this may be raw JSON, or a table, a graph, an image, a file path, or a 3D visualization.
+    3. Should easily differentiate between local and remote i/o.???
```
swashcap: Any idea how we can clarify this item?

swashcap:
One question on a pipeline list item. I’d say this is good-to-go otherwise.

design/architecture-requirements.md
rssk: 'Should have provisions for remote and local specifications.'

I think the original is poorly worded, i/o makes me think of actual data, not a specification for the data. In the actual specification, it'll look more like

```
  "local": {
  },
  "remote:" {
  }
```

Which I think is explanatory enough?

swashcap: Yeah, that makes perfect sense. I’ll update.

swashcap: Addressed in 292aef6502e3f374b68eebb5d2112e9e887ea2ec.

rssk (approved):

PR #239, opened 2017-07-11
Refactor/consolidate-computation-ducks

tkah:
* **Problem:**
   1. Computation related functionality is partitioned into a couple ducks files, which can be confusing.
* **Solution:**
   1. Combine computation-related functionality into the same file.
* **Tests:**


.eslintrc.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
     ecmaFeatures: {
-      jsx: true
+      jsx: true,
+      experimentalObjectRestSpread: true
```
swashcap: What's up with this new prop?

packages/coinstac-ui/app/render/state/root-reducer.js
```diff
@@ -12,8 +11,7 @@ import { reducer as form } from 'redux-form';
 
 const rootReducer = combineReducers({
   auth,
-  computation,
-  computations,
+  computationsState,
```
swashcap: I like the brevity of `computations`, but `computationsState` makes sense here. Wonder if there’s a way to easily refactor computation use into a single collection instead of the collection hanging from `computations` and the “selected” computation at `computation`. Our implementation doesn’t clearly separate entity state from UI state, which is unfortunate.

swashcap (approved):

swashcap:
@tkah 😕 Merge conflicts again. So many PRs!

.eslintrc.js
tkah: It allows the use of the `spread` operator in creating new objects without it getting pinged by the linter. I used it a fair amount in the reducers.

swashcap:
🎉 Great!

PR #243, opened 2017-07-13
Refactor/Getting results ducks in order

tkah:
* **Problem:**
   1. Results related functionality is partitioned into a couple ducks files, which can be confusing.
* **Solution:**
   1. Combine results-related functionality into the same file.
* **Tests:**


packages/coinstac-ui/app/render/state/ducks/results.js
```diff
@@ -0,0 +1,67 @@
...
+
+const INITIAL_STATE = {
+  localResults: [],
+  remoteResults: [],
+};
+
+// Reducer
+export default function reducer(state = INITIAL_STATE, action = {}) {
+  switch (action.type) {
+    case 'SET_LOCAL_RESULTS':
```
swashcap: These `case`s should be the constants themselves, not the strings.

packages/coinstac-ui/app/render/state/root-reducer.js
```diff
@@ -20,7 +20,7 @@ const rootReducer = combineReducers({
   loading,
   project,
   projects,
-  remoteResults,
+  resultsState,
```
swashcap: What would you think about naming this part of the state tree `results` instead of `resultsState`? Bit terser, and the keys are `localResults` and `remoteResults` so it shouldn't cause confusion (no `results.results` plucking).

swashcap (changes requested):

swashcap (approved):

PR #244, opened 2017-07-13
Enhancement/Every process needs a log

tkah:
* **Problem:**
   1. Main and render processes log their information to the same file, which can cause some confusion/inconvenience when debugging.
* **Solution:**
   1. Separate the `logger` into `mainLogger` and `renderLogger`, each pointing to different file.
* **Tests:**
   1. Check that two separate log files exist in the log directory.

### DIFF
--- a/packages/coinstac-client-core/src/index.js
+++ b/packages/coinstac-client-core/src/index.js
@@ -89,7 +89,7 @@ class CoinstacClient {
     this.appDirectory = opts.appDirectory ||
       CoinstacClient.getDefaultAppDirectory();
     this.halfpennyBaseUrl = opts.hp;
-    this.logger = opts.logger || new Logger({ transports: [new Console()] });
+    this.logger = opts.mainLogger || new Logger({ transports: [new Console()] });
 
     // hack for electron-remote. generate full API, even if it's dead.
     this.auth = {};
--- a/packages/coinstac-ui/app/main/utils/boot/configure-logger.js
+++ b/packages/coinstac-ui/app/main/utils/boot/configure-logger.js
@@ -15,8 +15,18 @@ module.exports = function configureLogger() {
 
   return mkdirp(logLocation, parseInt('0775', 8))
   .then(() => {
-    const logFilePath = path.join(logLocation, app.config.get('logFile'));
-    const logger = new winston.Logger({
+    let logFilePath = path.join(logLocation, app.config.get('mainLogFile'));
+    const mainLogger = new winston.Logger({
+      transports: [
+        new winston.transports.Console(),
+        new winston.transports.File({
+          filename: logFilePath,
+        }),
+      ],
+    });
+
+    logFilePath = path.join(logLocation, app.config.get('renderLogFile'));
+    const renderLogger = new winston.Logger({
       transports: [
         new winston.transports.Console(),
         new winston.transports.File({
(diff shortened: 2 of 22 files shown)


packages/coinstac-client-core/src/index.js
```diff
@@ -89,7 +89,7 @@ class CoinstacClient {
     this.appDirectory = opts.appDirectory ||
       CoinstacClient.getDefaultAppDirectory();
     this.halfpennyBaseUrl = opts.hp;
-    this.logger = opts.logger || new Logger({ transports: [new Console()] });
+    this.logger = opts.mainLogger || new Logger({ transports: [new Console()] });
```
swashcap, line 92: We should keep this as `logger` so it’s more general. _coinstac-client-core_ should be environment-agnostic – unaware that it's running in a “main” Electron process.

packages/coinstac-ui/app/render/state/ducks/util.js
```diff
@@ -1,7 +1,11 @@
 import app from 'ampersand-app';
 
+export const logUI = (level, msg) => {
+  app.renderLogger[level](msg);
```
swashcap, line 4: Why expose  the logger as a two-argument function as opposed the default level-methods? Calls look cleaner:

```js
import { logUI as log } from '../state/ducks/util';
// ...
log('Imma message');
log.error('Oh noes!');
```

…versus:

```js
import { logUI as log from '../state/ducks/util';
// ...
log('info', 'Imma message');
log('error', 'Oh noes!');
```

This also means that one must look for two cases when attempting to track down log calls (`error`, for example: `log\.error` and `log\('error'`).

packages/coinstac-ui/app/main/utils/boot/configure-logger.js
```diff
@@ -26,29 +36,39 @@ module.exports = function configureLogger() {
     });
 
     if (cliOpts.loglevel) {
```
swashcap, line 38: Not sure we ever use CLI-based log levels. Wonder if we need it.

packages/coinstac-ui/app/render/state/ducks/util.js
swashcap: Also, `logUI` is a bit redundant: we’re in the `render` directory, and these calls happen from modules containing user interface components. How does simply `log` sound?

packages/coinstac-ui/app/render/utils/boot/configure-logger.js
```diff
@@ -11,20 +12,29 @@ const mainLogger = electron.remote.require('app/main/utils/expose-app.js')().log
...
+function getLogger(consoleMethod, winstonMethod, logger) {
   return (...args) => {
     console[consoleMethod].apply(console, args); // eslint-disable-line no-console
-    mainLogger[winstonMethod].apply(null, ['ui', ...args]);
+    logger[winstonMethod].apply(null, ['ui', ...args]);
   };
 }
 
 module.exports = function configureLogger() {
-  app.logger = {
```
swashcap, line 22: Can’t remember why we’re doing this. Seems pretty convoluted: a separate logger (wrapper?) for every method? This essentially decorates Winston methods. Wonder if there's another way around it.

packages/coinstac-ui/app/render/state/ducks/util.js
swashcap: _Also_ also, the render process’s logger still lives in the main process, and log calls are synchronous over RPC. Wonder if there's a better approach for this.

swashcap (changes requested):

rssk:
What's up w/ this @tkah ?

PR #246, opened 2017-07-19
📄 Conditional pipeline flow

swashcap:
Add a note about conditional pipeline flow. Probably need some more info on this in the architecture doc.


swashcap:
@rssk Thoughts on this line? Can we merge?

rssk (approved):

PR #288, opened 2017-09-19
Fix login and allow new inspector debugging protocol

rssk:
# Problem(s) and fix(es)
The user model used for login was broken due to an upstream DB change,
this is fixed. Updated electron to use new inspector protocol for
debugging, this necessitated a change to the arg parsing.

To use the new inspector protocol add the following line to the relevant section in your VSCode `launch.json`

```
            "protocol": "inspector",
```
This also probably requires the latest version of VSCode as well.


tkah:
🎉

PR #298, opened 2017-09-26
Fixed Icons for dev

SwiftKiII:


rssk:
Hey Alex, looks like the relative path worked fine for both instances, so I just used that

PR #299, opened 2017-10-04
Redesign/consortia delete

SwiftKiII:
* **Problem:**
   1. Consortium items were able to be created but not deleted
* **Solution:**
   1. Created a graphql mutation to accommodate deleting a consortium item 
   1. Added code in resolvers.js to delete consortium item from rethinkdb
   1. Deleted cached consortium item so that the front end 
   1. User permissions determine front-end access and back-end GraphQL access.
* **Tests:**
   1. Click on consortia tab on the left
   1. Create new consortium if one with the delete button does not exist
   1. Click on delete button to confirm that the consortium was deleted


packages/coinstac-graphql/src/data/resolvers.js
```diff
@@ -137,7 +137,15 @@ const resolvers = {
       return new Promise();
     },
     deleteConsortiumById: (_, args) => {
-      return new Promise();
+      return helperFunctions.getRethinkConnection()
```
tkah: We should be watching permissions in our resolvers now, so I'd add a few lines here to limit deletion to only those with write access to the `consortia` table or write access to the submitted consortium id. Something like this:

```
deleteConsortiumById: ({ auth: { credentials: { permissions } } }, args) => {
      if (!permissions.consortia.write
          && args.consortium.id
          && !permissions.consortia[args.consortium.id].write) {
      }

      return helperFunctions.getRethinkConnection()
      ...

File: packages/coinstac-client-core/package.json [hunk 2]
```diff
@@ -38,7 +38,7 @@
     "coins-deposit-box": "^5.1.0",
     "coinstac-common": "^2.6.0",
     "coinstac-computation-registry": "^2.6.0",
-    "coinstac-computation-registry-new": "file:///Users/torrankahleck/coinstac/packages/coinstac-computation-registry-new",
+    "coinstac-computation-registry-new": "file:///Users/aramirez/coinstac/packages/coinstac-computation-registry-new",
```
tkah: Once this is merged, we shouldn't have to deal with this mess anymore.

packages/coinstac-ui/app/render/components/consortium/consortia-list-item.jsx
```diff
@@ -1,15 +1,18 @@
...
 
-const ConsortiaListItem = ({ owner, user, consortium }) => (
+const ConsortiaListItem = ({ owner, user, consortium, submit}) => (
   <Panel header={<h3>{consortium.name}</h3>}>
     <p>{consortium.description}</p>
     <LinkContainer to={`/consortia/${consortium.id}`}>
       <Button bsStyle="info">View Details</Button>
     </LinkContainer>
-    {owner && <Button bsStyle="danger" className="pull-right">Delete Consortium</Button>}
+    {owner && <Button bsStyle="danger" onClick={() => submit(`${consortium.id}`)} className="pull-right">Delete Consortium</Button>}
```
tkah: When your lines start looking a bit long, consider breaking them down:

```
{owner &&
  <Button
    bsStyle="danger"
    onClick={() => submit(`${consortium.id}`)}
    className="pull-right"
  >
      Delete Consortium
  </Button>
}

Comment (tkah) [inline packages/coinstac-ui/app/render/components/consortium/consortia-list-item.jsx]: Sorry to get nitpicky, but would you consider renaming this function from `submit` to `deleteConsortium` or something more relevant to its purpose?

File: packages/coinstac-ui/app/render/components/consortium/consortia-list-item.jsx [hunk 4]
```diff
@@ -19,6 +22,23 @@ ConsortiaListItem.propTypes = {
...
+    submit: (consortiumId) => mutate({
+      variables: { consortiumId },
+      update: (store, { data: { deleteConsortiumById } }) => {
+        const data = store.readQuery({ query: fetchAllConsortiaFunc });
+        const index = data.fetchAllConsortia.findIndex(con => con.id === deleteConsortiumById.id);
+        if (index > -1) {
+          data.fetchAllConsortia.splice(index, 1);
+        }
+        store.writeQuery({ query: fetchAllConsortiaFunc, data });
+      }
```
tkah: 🥇

tkah (changes requested):
Looks good. Just a few minor formatting/naming concerns. 👍

tkah (approved):

PR #302, opened 2017-10-10
Redesign/consortium save fix

SwiftKiII:
* **Problem:**
   1. Consortium items were able to be created but not edited
* **Solution:**
   1. Added code in resolvers.js to save consortium item from rethinkdb to both accurately save information as well as updated the users tables based on permissions for the new consortium. This allows the user to later edit and save changes
* **Tests:**
   1. Click on consortia tab on the left
   1. Create new consortium or click on view details of an already made consortium
   1. Make any changes you desire to name and description and hit save
   1. Navigate back to consortium tab and see updated consortium name/description


packages/coinstac-graphql/src/data/resolvers.js
```diff
@@ -123,15 +123,20 @@ const resolvers = {
...
+              conflict: "update",
               returnChanges: true,
             }
           )
           .run(connection)
+          .then((result) => {
+            //r.db('coinstac').table('users').filter({id: 'test'}).update({'permissions': {'consortia': {'some-new-id': {'write': true}}}})
+            let crazything = {};
+            crazything[result.changes[0].new_val.id] = {'write': true};
+            rethink.table('users').filter({id: 'test'}).update({'permissions': {'consortia': crazything}})
```
tkah: 😏 Should probably use something a bit more descriptive instead of `crazything`. You could actually use `consortia` as the name and then use [ES6 Property Value Shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer) since the property and the value would have the same name:

[(Another Example)](https://github.com/airbnb/javascript#es6-object-concise)

```
  let consortia = {};
  consortia[result.changes[0].new_val.id] = {'write': true};
  rethink.table('users').filter({id: 'test'}).update({'permissions': { consortia }})
```

Also, watch your spacing on objects. MRN uses the [AirBnB styleguide](https://github.com/airbnb/javascript), which [recommends padding when using curly braces](https://github.com/airbnb/javascript#whitespace--in-braces). The linter should generally warn you about this, and it will fail once we get testing running again so it's not a bad habit to get into now.

tkah (changes requested):

tkah:
Looks good! Just a couple small things.

packages/coinstac-graphql/src/data/resolvers.js
tkah: Doesn't seem like we should be hardcoding the `id` here... You should be able to get the user's id from the auth parameter. Just have the change the function a bit:

```
saveConsortium: ({ auth: { credentials } }, args) => {
  const { permissions } = credentials;
  if (!permissions.consortia.write
          && args.consortium.id
          && !permissions.consortia[args.consortium.id].write) {
      }
  ...
  rethink.table('users').filter({ id: credentials.id }).update({'permissions': { consortia }})
```

PR #313, opened 2017-11-08
Redesign/pipeline save

SwiftKiII:
* **Problem:**
   1. Consortia and Pipeline pages functionality added. Details listed below.
* **Pipeline page:**
   1. Added ability to save any changes to DB (will only allow user to save if any changes were made).
   1. Cleaned up UI.
   1. Added owning consortium field on UI and schema (defaults current owning consortium).
   1. Added ability to delete entire computations
   1. Added ability to see all pipelines and edit pipelines that have already been created.
* **Consortia page:**
   1. Added ability to set, change, and view activepipeline
   1. Can route to "create pipeline page" from consortia page that auto-populates  owning consortium field.
   1. Cleaned up UI
   1. Added ability to select both owned pipelines and shared pipelines


packages/coinstac-ui/app/render/components/consortium/consortium-pipeline.jsx
```diff
@@ -1,53 +1,106 @@
...
-    <div>
+class ConsortiumPipeline extends Component{
+  constructor(props) {
+    super(props);
+
+    let ownedPipelines = null;
+    let sharedPipelines = null;
+    let activePipelineRef = {};
+
+    const data = ApolloClient.readQuery({ query: fetchAllPipelinesFunc });
```
tkah: Perhaps this is why you loaded the fetchAllPipelinesFunc on the list page? The problem with that is that a person could conceivably hit the refresh button on this page and not have the fetchAllPipelinesFunc preloaded anymore to pull from as this is on a completely different route. I'd suggest either creating separate Apollo queries for `ownedPipelines` and `sharedPipelines` or moving the `fetchAllPipelines` call to `consortium-tabs.jsx` since that is the parent of this same route. However, the problem I have with the latter is that I think we'll eventually make all of these tabs their own route, meaning that we'll end up having to refactor it then anyway to load the queries here. Also, maybe it won't finish the query before loading its subcomponents anyway?

packages/coinstac-ui/app/render/components/consortium/consortium-tabs.jsx
```diff
@@ -24,7 +24,8 @@ class ConsortiumTabs extends Component {
       description: '',
       users: [],
       owners: [],
-      activeComputationId: '',
+      activePipeline: '',
+      activeComputationId:'',
```
tkah: I'm gonna go through and remove all the references to `activeComputationId` along with the merge. It was a mistake.

packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
```diff
@@ -43,23 +49,49 @@ class Pipeline extends Component {
...
+      consortium: consortium,
+      startingPipeline: pipeline,
     };
 
     this.addStep = this.addStep.bind(this);
     this.moveStep = this.moveStep.bind(this);
     this.updateStep = this.updateStep.bind(this);
+    this.deleteStep = this.deleteStep.bind(this);
+    this.updatePipeline = this.updatePipeline.bind(this);
+    this.open = this.open.bind(this);
```
tkah: Could these names be a bit more descriptive? We also have accordions on the page, and I thought that was what this referred to at first. `openModal()`?

packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
```diff
@@ -194,30 +318,88 @@ class Pipeline extends Component {
...
+          </Row>
+
+          <Row>
+            <Col sm={12}>
+              {consortia &&
+              <SplitButton 
+                bsStyle="info" 
+                title={consortium ? consortium.name : "Consortia"} 
+                id="pipelineconsortia" 
+                onSelect={(value, e) => this.updatePipeline({ param: 'owningConsortium', value: value, consortiumName: e.target.innerHTML })}>
```
tkah: I found a better way to write fewer anonymous functions like this by using a closure in the function instead. Change `updatePipeline` to the following:

```
updatePipeline(update) {
  return (...args) {
    let target = null;
    let eventKey = '';
    
    if (args.length === 1) {
      target = args[0].target;
    } else {
      eventKey = args[0];
      target = args[1].target;
    }

    if(update.param == 'owningConsortium') {
      this.setState({ consortium: { id: eventKey, name: target.innerHTML }});
    } else {
      this.setState(prevState => ({
        pipeline: {
          ...prevState.pipeline,
          [update.param]: target.checked !== undefined ? target.checked : target.value,
        },
      }));
    }
  }
}
```

And call as follows: 
`onChange={this.updatePipeline({ param: 'shared' })}`

The concern is that anonymous functions in the `render()` method are recreated every time `render()` is called, creating more overhead. Using closures in this way gets around that issue.

In fact, I'm looking at refactoring even further, starting with new component functions I write. It seems you can automatically bind functions to `this` by using arrow functions in their definitions and skip binding them in the constructor. See [here](https://github.com/facebook/react/issues/9851) and [here](https://medium.com/@machnicki/handle-events-in-react-with-arrow-functions-ede88184bbb). But I don't think that's important now as all of the other components are written the way you have it currently.

packages/coinstac-ui/app/render/components/consortium/consortium-pipeline.jsx
```diff
@@ -1,53 +1,106 @@
...
+    ownedPipelines = data.fetchAllPipelines.filter(pipe => pipe.owningConsortium === props.consortium.id);
+    sharedPipelines = data.fetchAllPipelines.filter(pipe => pipe.shared);
+
+    if(this.props.consortium.activePipeline.length){
```
tkah: This throws an error for me when I try to visit the page. Something about _can't read property length of null_. I don't have an `activePipeline` on my consortium object, so there's nothing to test against.

Try changing it to: `if (props.consortium.activePipeline) {` or even `if (props.consortium.activePipeline && props.consortium.activePipeline.length) {` if there are times when the field will exist but only consist of an empty string.

tkah (changes requested):
Just a few changes from looking through the code and starting it up. However, make sure you're running the linter before you commit. There are a few linting errors you need to take care of. It should run on its own now if you stop typing `get commit -n -m ""`. You don't need the `-n` anymore. The broken tests have been removed.

Also, we need to make sure your live linting is working in VSCode as there are a bunch more in there that aren't tripping up the precommit linter. For instance, `consortium-pipeline.jsx` has quite a few minor formatting issues.

packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
```diff
@@ -179,10 +211,102 @@ class Pipeline extends Component {
...
+    this.setState({ 
+      showModal: true,
+      stepToDelete: stepId,
+    });
+  }
+
+  updatePipeline(update) {
+    if(update.param == 'owningConsortium')
+      this.setState({ consortium: { id: update.value, name: update.consortiumName }});
+    this.setState(prevState => ({
```
tkah: Might want to wrap this in an `else`?

packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
tkah: After looking at this more, I'd be fine if you didn't make this change. Seems needlessly complex. Though we probably should generally shoot for using closures over anonymous functions.

tkah (approved):

PR #344, opened 2018-01-11
Redesign/electron beta node8

tkah:
* **Problem:**
   1. Non-owners can move around pipeline steps. Back-end permissions will prevent them from saving the step reordering, but allowing non-owners to move them is misleading.
* **Solution:**
   1. Minor check of owner on the moveable steps.
* **Tests:**
   1. Login as the `test` user and attempt to move the steps in the test pipeline to see that you aren't allowed.


tkah:
Incorrect branch

PR #373, opened 2018-02-12
Local Results, Schema Changes and IPCRenderer Fix

### LINKED ISSUE #369: Fix UI inputMap Schema to Match Pipeline inputMap Schema
# Problem
The expected pipeline schema from `coinstac-pipeline` doesn't match the pipeline schema outputted by the UI.
# Solution
- [x] Covariates field should not be assumed.
- [x] All inputs should have a possible source field, not just covariates.
- [x] All inputs with input sources mapped to previous pipeline computation outputs should automatically remap their inputs when a computation step changes position in the overall pipeline.
- [x] Covariates need to function user-side as before with new schema changes.
- [x] Collection pipelines inputs with user file sources need to work with schema changes.
  - [x] In activeConsortia, insert null into stepIO indexes so all indices are matched with respective pipelineSteps inputs/covars
  - [x] Make collection fileGroups an object with ids as params (?)

### LINKED ISSUE #372: Multiple Subscriptions on UI Refresh?
# Problem
There are signs that subscriptions might be doubled-up when the UI refreshes versus stopping and then starting:
- After a dozen or so refreshes, warnings appear concerning too many listeners.
- One *computations downloaded* notification appears for each time the app has refreshed. 
# Solution
Investigate cause for above oddities.

tkah:
* **Problem:**
   1. Local runs not included in results displays.
   1. All controllers `local`.
   1. Closes #372 & closes #369
* **Solution:**
   1. Group GraphQL `remoteRuns` and IndexDB `localRuns` into Redux `runs`.
   1. Controller types now dependent upon presence of *remote* section in computation schema.
   1. Remove auto-inclusion of `covariates` field in pipeline `inputMap`.
   1. All inputs have an owner defined source field.
   1. All inputs, including nested covars, will auto-remap when possible on step position changes.
   1. Collection `fileGroups` made into object to make matching pipeline step output easier.
   1. Empty arrays added to `associatedCollection` to make mapping file output easier.
* **Tests:**
   1. Login and refresh UI.
   1. Pull down computation (remove one first if all already downloaded).
   1. Download complete notification should only display once.
   1. Make **Local Pipeline** active in *Test Consortia* and run it.
   1. Remote server should never be called and results should still be listed in results page.
   1. Other changes should not change functionality.


packages/coinstac-api-server/test/test-setup.js
```diff
@@ -100,6 +100,35 @@ helperFunctions.getRethinkConnection()
...
+        },
+      },
+      {
+        id: 'UIKDl-local2',
+        controller: { type: 'local' },
+        computations: [
+          compInsertResult.changes[0].new_val.id,
+        ],
+        inputMap: {
+          start: { fromCache: { step: 0, variable: 'sum', label: 'Sum' } },
```
rssk: While this works, I'd prefer to reference the compspec for the Label attribute

packages/coinstac-ui/app/render/components/consortia/consortia-list.jsx
```diff
@@ -195,8 +200,54 @@ class ConsortiaList extends Component {
...
-  startPipeline(consortiumId) {
+  startPipeline(consortiumId, activePipelineId) {
     return () => {
+      const { client } = this.props;
+      let isLocalPipeline = false;
+      const pipelineData = client.readQuery({ query: FETCH_ALL_PIPELINES_QUERY });
+      const pipeline = pipelineData.fetchAllPipelines
+        .find(pipe => pipe.id === activePipelineId);
+
+      for (let i = 0; i < pipeline.steps.length; i += 1) {
```
rssk, line 211: I think it might make more sense if the last step isn't `remote`, as you could have several local preproc steps, and the final output is what matters for now

packages/coinstac-ui/app/render/components/user/user-account-controller.jsx
```diff
@@ -25,6 +25,7 @@ class UserAccountController extends Component {
       });
     })
     .catch((err) => {
+      console.log(err);
```
rssk: ?

rssk (changes requested):

packages/coinstac-ui/app/render/components/consortia/consortia-list.jsx
tkah: This is backwards and is fixed in #380 to look for the existence of a `decentralized` step, which will flip the flag, `isRemotePipleine`, to true and break. It's not used to determine anything concerning the output. It's just used to determine whether or not to send the run to GraphQL and the remote server, or to keep things limited to the local machine.

packages/coinstac-ui/app/render/components/user/user-account-controller.jsx
tkah: Yeah, this can be taken out. Just a bit quicker for me to reference the console output rather than the log.

packages/coinstac-api-server/test/test-setup.js
tkah: That's fine. I'll take it out of the test file here and ensure it's looking to Computation on the branch I'm working on now.

rssk (approved):

PR #390, opened 2018-02-19
Can view pipelinesnapshots on pipeline page

### LINKED ISSUE #385: Pipeline Page Works With Run Pipeline Snapshots
# Problem
There's no convenient way to view the pipeline steps involved in a past run.
# Solution
Modify the pipeline component to be able to load run pipeline snapshots.

SwiftKiII:
close #385

Can now view pipelinesnapshots on pipeline page. The user will not be able to change any data as it is a reference to the pipeline as it was when the run was created. In this PR I also have UI visual changes to results-list-item due to feedback.

### DIFF
--- a/packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
+++ b/packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
@@ -54,7 +54,7 @@ class Pipeline extends Component {
     super(props);
 
     let consortium = null;
-    const pipeline = {
+    let pipeline = {
       name: '',
       description: '',
       owningConsortium: '',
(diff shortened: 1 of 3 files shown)


packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
```diff
@@ -109,9 +115,13 @@ class Pipeline extends Component {
 
   setConsortium() {
     const { auth: { user }, client } = this.props;
-    const owner = user.permissions.consortia[this.state.pipeline.owningConsortium] &&
+    let owner;
+    if (this.props.params.runId) {
+      owner = false;
```
tkah, line 120: Nice. Seems pretty simple.

tkah (approved):
Nice and clean

PR #392, opened 2018-02-20
Pipeline: Bug/remove remove mapping

rssk:
# Problem
Remote could possibly break on a unfulfilled spec from an Owner pipeline start, but remotes dont really use the input map in specs anyway....

# Solution
Remote no longer use given inputmaps


tkah (approved):
Looks good

PR #402, opened 2018-02-27
Results are now displayed with React-Faux-Dom

### LINKED ISSUE #386: Table Display
# Problem
There is no table generic display for results
# Solution
Create a table display for data in results page

### LINKED ISSUE #389: Convert scatter and box plot to react-faux-dom
#Problem 
Current d3 react libraries are complex and dont support every type of chart or image we need.

#Solution 
Convert current d3 libraries to react-faux-dom

SwiftKiII:
# Problem
referenced issue(s):
close #386
close #389 
# Solution
Converted all charts to use React-Faux-Dom with new D3 implementations. Also added a table view for results.
# Testing
Changed between the different results listings (Scatter, Boxplot, Table)


packages/coinstac-ui/package.json
```diff
@@ -38,6 +39,7 @@
     "react-dnd": "^2.5.4",
```
tkah: Could you get rid of these unused d3 libraries?

packages/coinstac-ui/app/render/components/results/result.jsx
```diff
@@ -41,46 +52,353 @@ class Result extends Component {
       });
   }
 
+  drawScatter() {
```
tkah: I know it's a bit of a hassle, but it'll be much more manageable going forward if we move the various plots into their own components and just import them into `result.jsx`. Maybe we could make a subdirectory in `results` called `graphs` or `plots`? Then put `box-plot.jsx`, `scatter-plot.jsx` and `results-table.jsx` components in there?

tkah (changes requested):
Some minor refactoring. Graphs look good.

packages/coinstac-ui/package.json
tkah, line 35: Could you get rid of some of these unused d3 libs?

tkah (approved):

tkah:
Tabs look good.

PR #406, opened 2018-02-28
Display Unmapped Inputs Warning

### LINKED ISSUE #388: Validate Pipeline Owner Mappings to Enable Start Button
# Problem
At no point are pipelines checked to have have all fields mapped.
# Solution
Owner-side pipeline mapping should be checked before a pipeline is allowed to start.

### LINKED ISSUE #400: Pipeline Changes Don't Propagate to User LocalDB
# Problem
* Changes to pipeline won’t change pipelineSteps in associatedConsortia
# Tasks
* Swap out pipelineSteps for new steps on change
* Delete `stepIO` in associatedConsortia

### LINKED ISSUE #401: Remote Consortium Changes to Pipeline Not Reflected Locally
# Problem
Changes to active pipeline handled on owner machine but not clients.
# Tasks
Detect changes to remote consortium active pipeline id and change local activeConsortia details.

tkah:
# Problem
- No warning given to owners prior to start pipeline button that their mappings are incomplete
- Closes #401 
- Closes #400 
- Closes #388 
# Solution
- Use remote pipeline and consortia changes to investigate if pipeline mappings are complete and reset them if not
- Mapping complete check inserted after each var mapped in collections
- Warnings inserted on *Consortia List* if either no `activePipelineId` is set or if client-side mappings are incomplete.
# Testing
1. Test pipelines don't require any client-side mappings. Ensure the **Start Pipeline** button is enabled when either `local` or `decentralized` pipelines are set to active on *Test Consortium 2*
1. *Test Pipeline 1* does require client side mappings. Set this as active on *Test Consortium 2* and ensure **Start Pipeline** button is replaced by warning.
   1. Mapped inputs to *Test Pipeline 1* and ensure that the **Start Pipeline** has reappeared.


packages/coinstac-api-server/test/data/multi-shot-schema.json
```diff
@@ -33,14 +34,20 @@
...
+        "source": "owner"
       },
       "covariates":
       {
         "defaultValue": "[{'name': 'Is Control?','type': 'boolean'}]",
         "label": "Covariates",
         "type": "array",
         "items": ["boolean", "number"]
+      },
+      "data": {
```
rssk, line 47: Would type be `files` here? Then the schema would let you put on a extension to test (when we do that)

packages/coinstac-api-server/test/test-setup.js
```diff
@@ -67,14 +68,13 @@ helperFunctions.getRethinkConnection()
...
-              name: 'biased xs',
-              source: { inputKey: 'biasedX', inputLabel: 'Computation 1: Biased Xs', pipelineIndex: 0 },
-              type: 'number',
-            },
-          ],
-          freeSurferRegion: { value: ['5th-Ventricle'] },
+          covariates: {
+            ownerMappings: [
+              {
+                fromCache: { variable: 'biasedX', step: 0 },
```
rssk, line 74: Interesting, I didn't quite think of this and wonder if it'll be used, but see no harm in it.

rssk (changes requested):

packages/coinstac-api-server/test/test-setup.js
tkah: Yeah, for now it just shows the link in the UI, but the case isn't really handled when data is passed off to pipeline. I figured I'd leave it there as an idea for now if we want to return to it.

packages/coinstac-api-server/test/data/multi-shot-schema.json
tkah: I think another field would be better to use for that. Right now, the `array` type informs the UI how to display the input field. And since the actual data structure is an array, I think it fits well.

rssk (approved):

PR #436, opened 2018-03-15
Results back button

### LINKED ISSUE #384: Results Back Button
# Problem
There are multiple ways of accessing results. But no way to get to directly return to the screen that brought you there.
# Solution
Create a back button to take the user back to whichever component preceded their arrival at a given result details page.

### LINKED ISSUE #426: UI: Show Numerical Table Display Data with Scientific Notation
# Problem
On the table results page, the numbers are displayed with all digits. I recommend using scientific notation for numbers smaller than 0.001 and greater than 999. I recommend displaying five significant digits.

### LINKED ISSUE #427: UI: Display Table Results Fields Containing Arrays as Arrays
# Problem
On the results table page, it seems that the array elements are concatenated together into one string. They should be displayed as separate elements.

SwiftKiII:
# Problem
referenced issue(s):
closes #384 
closes #426 
closes #427 

# Solution
Made a few suggested visual changes to how the data is being displayed
# Testing


tkah:
Looks good

PR #443, opened 2018-03-20
Error on Non-existent Docker Hub Images

### LINKED ISSUE #416: Catch Errors When Pulling Non-existent Docker Images
# Problem
If a Docker image doesn't exist, comp-registry throws a nasty error and kills the call
# Tasks
Handle non-existent docker images

### LINKED ISSUE #434: Server: comp reg not reliably pulling in images
# Problem
upon updating the new `coinstac/ssr` image in ssr_test the reg did not repull/dl the image
# Tasks
- just do it :tm:

tkah:
# Problem
* Closes #434 
* Closes #416 
# Solution
* Uncomment automatic image downloading on server start
* Uncomment image downloading via UI
* Pass on error if image doesn't exist in Docker Hub
# Testing
* Starting `coinstac-server` should download all images and show an error message for any that don't exist in Docker Hub
* Try to download *Multi-Shot Regression* via the UI and observe the error message.

### DIFF
--- a/packages/coinstac-docker-manager/src/index.js
+++ b/packages/coinstac-docker-manager/src/index.js
@@ -195,10 +195,10 @@ const startService = (serviceId, serviceUserId, opts) => {
  * @return {Object} Returns stream of docker pull output
  */
 const pullImage = (computation) => {
-  return new Promise((resolve, reject) => {
+  return new Promise((resolve) => {
     docker.pull(computation, (err, stream) => {
       if (err) {
-        reject(err);
+        resolve(err);
       }
       resolve(stream);
     });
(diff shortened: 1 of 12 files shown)


packages/coinstac-docker-manager/src/index.js
```diff
@@ -195,10 +195,10 @@ const startService = (serviceId, serviceUserId, opts) => {
  * @return {Object} Returns stream of docker pull output
  */
 const pullImage = (computation) => {
-  return new Promise((resolve, reject) => {
+  return new Promise((resolve) => {
     docker.pull(computation, (err, stream) => {
       if (err) {
-        reject(err);
```
rssk: So I don't particularly agree with this, I think the lower level function should reject if you have an image that is invalid, otherwise the behavior is inconsistent. _However_ the real solution is in comp registry to use `Promise.all` in a way to proceed even when one errors https://davidwalsh.name/promises-results.

You'll do that, then also pass an array of errors (if some reject) and results, and then deal w/ or display that.

rssk (changes requested):

PR #478, opened 2018-04-10
UI: Snackbar on Consortia Save, Computation Creation, and Collection Save

### LINKED ISSUE #420: UI: Snackbar on consortia save
# Problem
* Notification appearance is a bit lame.
* There isn't any feedback on saving consortia or pipelines or collections.
# Tasks
* Update the look of all notifications to look more similar to Material style [snackbars](https://www.material-ui.com/#/components/snackbar) using the current library [react-notification-system-redux](https://github.com/gor181/react-notification-system-redux). Keep the location in the upper-right-hand corner for now.
* Call existing notification functions (`packages/coinstac-ui/app/render/state/ducks/notifyAndLog.js`) on save of consortia, pipelines, and collections.

pixelsaurus:
# Problem

- Notification appearance is a bit lame.
- There isn't any feedback on saving consortia or pipelines or collections.
- Closes #420

# Solution

- Update the look of all notifications to look more similar to Material style snackbars using the current library react-notification-system-redux. Keep the location in the upper-right-hand corner for now. 
- Call existing notification functions (packages/coinstac-ui/app/render/state/ducks/notifyAndLog.js) on save of consortia, pipelines, and collections.

# Testing

- Create a Consortium
- Create a Computation
- Create a File Collection > Save under About Tab

### DIFF
--- a/packages/coinstac-ui/app/render/styles/alert.scss
+++ b/packages/coinstac-ui/app/render/styles/alert.scss
@@ -16,16 +16,3 @@
     border: none;
   }
 }
-
-/**
- * Notifications.
- *
- * Override react-notification-system-redux component styles.
- */
-
-.notification {
-  border-radius: 0 !important;
-  border: none !important;
-  opacity: 0.75 !important;
-  box-shadow: none !important;
-}
--- a/packages/coinstac-ui/app/render/styles/alert.scss
+++ b/packages/coinstac-ui/app/render/styles/alert.scss
@@ -17,3 +17,15 @@
   }
 }
 
+/**
+ * Notifications.
+ *
+ * Override react-notification-system-redux component styles.
+ */
+
+.notification {
+  border-radius: 0 !important;
+  border: none !important;
+  opacity: 0.75 !important;
+  box-shadow: none !important;
+}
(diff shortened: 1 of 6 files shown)


packages/coinstac-ui/app/render/styles/alert.scss
```diff
@@ -17,3 +17,15 @@
...
+ *
+ * Override react-notification-system-redux component styles.
+ */
+
+.notification {
+  border-radius: 0 !important;
+  border: none !important;
+  opacity: 0.75 !important;
+  box-shadow: none !important;
+}
```
tkah: This works, but the notification looks a little jerky on my machine. Might have to do with recently having built and the thing being a little overworked. Anyway, what doesn't cause it to be a little jerky, and will also allow us to avoid the dreaded `!important`, is using the component's [built-in way for overriding styles](https://github.com/gor181/react-notification-system-redux#usage).

In `coinstac-ui/app/render/components/app.jsx` add the following to the top just above the class declaration:
```js
const styles = {
  notifications: {
    NotificationItem: {
      DefaultStyle: {
        borderRadius: 0,
        border: 'none',
        opacity: 0.75,
        boxShadow: 'none',
        fontWeight: 'bold', // This might not be necessary. Use your judgement.
      },
    }
  }
};
```

Then change the *Notifications* component around line 49 to be something like:

```html
<Notifications
  notifications={notifications}
  style={styles.notifications}
/>
```

This will allow you to get rid of the scss here and, according to the docs, modify the styles of the various notification types should you want to some day.

pixelsaurus: Done :)

PR #501, opened 2018-05-01
Fixing csv parse matching, which is still and issue and bad idea as w…

rssk:
…ell fixes file link proms

# Problem
See title
# Solution

# Testing


packages/coinstac-ui/app/render/components/dashboard/dashboard.jsx
```diff
@@ -198,28 +198,24 @@ class Dashboard extends Component {
                 },
               });
 
-              // 5 second timeout to ensure no port conflicts in
```
rssk, line 201: no longer needed due to recursive port finding by docker-manager

packages/coinstac-ui/app/render/state/ducks/collections.js
```diff
@@ -58,7 +58,9 @@ function iteratePipelineSteps(consortium, filesByGroup, baseDirectory) {
                   mapIndex < step.inputMap[key].ownerMappings.length;
```
rssk, line 58: iteratePipelineSteps really just needs to be scrapped

PR #532, opened 2018-05-22
Pipeline: any node errorring now causes global error

### LINKED ISSUE #531: Pipeline
# Problem
Errors need to stop the pipe
# Tasks

rssk:
# Problem
referenced issue(s):
Closes #531

Pipeline errors would leave hanging pipelines, which is bad because that means they'll go forever...

# Solution
* make errors crash the pipeline, this means a local node crashes the whole thing, and joining nodes are notified if the pipe has already crashed.
* also tosses input data onto the error to help debugging

# Testing
`npm run test`


packages/coinstac-ui/app/main/index.js
```diff
@@ -131,7 +131,10 @@ loadConfig()
         core.unlinkFiles(run.id);
         mainWindow.webContents.send('local-run-error', {
           consName: consortium.name,
-          run: Object.assign(run, { error, endDate: Date.now() }),
+          run: Object.assign(
```
rssk, line 134: this was causing an RPC issue where the error wasnt being sent, I'm guessing there's an issue rpc'ing the whole Error proto stuff

pixelsaurus (approved):

PR #565, opened 2018-07-12
Pipeline: allow users to join early, fixing remote late starting issues

### LINKED ISSUE #549: Pipeline: docker downloads cause pipeline to not start
# Problem
If a download takes longer than any client, the run wont start as it loses that data
# Tasks

rssk:
# Problem
referenced issue(s):
Closes #549 

If the pipe hasnt called `start` the pipeline will hang
# Solution
This is due to there being no pipeline object to put data to, the solution is allowing temporary pipelines if there is none, waiting for the incoming start call.

# Testing
`npm run test`

### DIFF
--- a/packages/coinstac-pipeline/src/controller.js
+++ b/packages/coinstac-pipeline/src/controller.js
@@ -63,7 +63,6 @@ module.exports = {
        * @return {Promise}                resolves to the final computation output
        */
       start: (input, remoteHandler) => {
-        const queue = [];
         setStateProp('state', 'started');
         controllerState.remoteInitial = controllerState.mode === 'remote' ? true : undefined;
         if (!controllerState.initialized) {
@@ -155,7 +154,6 @@ module.exports = {
               throw new Error('unknown controller runType');
           }
         };
-        queue.push(iterateComp);
 
         /**
          * The trampoline works by continuously bouncing calls on and off the stack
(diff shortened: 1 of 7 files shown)


packages/coinstac-pipeline/src/controller.js
```diff
@@ -180,18 +178,21 @@ module.exports = {
          * with the trampoline this allows an unlimited number of steps without stack issues.
          * Credit to Dave's blog: http://www.datchley.name/asynchronous-in-the-browser/
          * @param  {Object}   initialInput the object to pass to computation
-         * @param  {Array}    steps        the dynamic array for the waterfall
          * @param  {Function} done         callback
          * @return {Object}                computation's result
          */
-        const waterfall = (initialInput, steps, done, err) => {
+        const waterfall = (initialInput, done, err) => {
+          const queue = [];
```
rssk, line 185: just bringing this down to better scope, was too high

packages/coinstac-pipeline/src/pipeline-manager.js
```diff
@@ -42,11 +42,12 @@ module.exports = {
 
     const waitingOnForRun = (runId) => {
       const waiters = [];
-      for (let [key, val] of Object.entries(remoteClients)) { // eslint-disable-line no-restricted-syntax, max-len, prefer-const
-        if (val[runId] && !val[runId].currentOutput) {
-          waiters.push(key);
+      activePipelines[runId].clients.forEach((client) => {
```
rssk, line 45: use the actual client list as the better source of truth

packages/coinstac-pipeline/src/pipeline-manager.js
```diff
@@ -277,8 +320,6 @@ module.exports = {
           throw new Error(`Unable to create pipeline directories: ${err}`);
         })
         .then(() => {
-          activePipelines[runId].state = 'running';
```
rssk, line 280: pipeline-manager pipeline state is actually not used anywhere, the UI state comes from the controller. Even so I thought it was useful to setup some state for the future, unfortunately this didnt work with everywhere with the `pre-pipeline` state solution /shrug

PR #581, opened 2018-10-04
Build script `platform` bug

swashcap:
# Problem

The `platform` variable in _scripts/build.js_ has a bad assignment.

referenced issue(s): _n/a_

# Solution

* Fix variable assignment
* Script refactor to clean it up

# Testing

```shell
npm run build
```


packages/coinstac-ui/scripts/build.js
```diff
@@ -6,12 +6,14 @@ const rm = pify(require('rimraf'));
 const os = require('os');
 const path = require('path');
 
-const platform = ['linux', 'win32', 'darwin'];
```
swashcap, line 9: The equality checks for icon compare an array to strings using strict equality. _WAT._

packages/coinstac-ui/scripts/build.js
```diff
@@ -6,12 +6,14 @@ const rm = pify(require('rimraf'));
...
+const platform = os.platform();
 const options = {
+  arch: os.arch(),
```
swashcap: Move this stuff up here so it's easier to see how `options` is set.

packages/coinstac-ui/scripts/build.js
```diff
@@ -50,5 +48,5 @@ rm('coinstac-*')
   });
 })
 .catch((err) => {
-  console.log(`Build failed with: ${err}`); // eslint-disable-line no-console
```
swashcap, line 53: Using string literals seems to serialize the error as `err.toString()`. This isn't so useful when debugging build errors.

packages/coinstac-ui/scripts/build.js
```diff
@@ -50,5 +48,5 @@ rm('coinstac-*')
...
+  console.error('Build failed with:', err); // eslint-disable-line no-console
```
swashcap, line 41: This outputs to stderr and ensures that the error's stack trace is output.

swashcap:
Bleh, looks like electron-packager supports and array of strings for the `platform` option: https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#platform. This isn't a bug with `platform` assignment! Going to address…

swashcap:
Hah: 615a5ef79a88efad916fb1d46456df1fac5735e7. It me.

swashcap:
I _think_ 2cee50b5764b3c06caebfaa6b13cd0628a373f8f addressed this. According to [electron-packager's 8.7.2 docs](https://github.com/electron-userland/electron-packager/blob/v8.7.2/docs/api.md) this should be valid.

PR #582, opened 2018-10-05
refactor: move UI artifacts to build directory

swashcap:
# Problem

* Placing the UI's built render file inside _app_ is annoying: it doesn't separate build artifacts from code, makes grep-ing for stuff difficult
* Dumping all the built apps in the _coinstac-ui_ root: same deal

referenced issue(s): _n/a_

# Solution

Move all artifacts/built apps to _packages/coinstac-ui/build_

# Testing

`npm run build`

### DIFF
--- a/packages/coinstac-ui/scripts/build.js
+++ b/packages/coinstac-ui/scripts/build.js
@@ -6,11 +6,12 @@ const rm = pify(require('rimraf'));
 const os = require('os');
 const path = require('path');
 
-const platform = ['linux', 'win32', 'darwin'];
+const platform = 'darwin'; // ['linux', 'win32', 'darwin'];
 const options = {
   asar: true,
   dir: `${__dirname}/../`,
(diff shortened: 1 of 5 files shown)


packages/coinstac-ui/scripts/build.js
```diff
@@ -6,11 +6,12 @@ const rm = pify(require('rimraf'));
 const os = require('os');
 const path = require('path');
 
-const platform = ['linux', 'win32', 'darwin'];
+const platform = 'darwin'; // ['linux', 'win32', 'darwin'];
```
rssk: ruh oh

rssk (approved):

packages/coinstac-ui/scripts/build.js
swashcap: Haha, whoops! Nice catch.

PR #583, opened 2018-10-05
Enhancement/fix linting errors

swashcap:
# Problem

Linting fails! :scream:

referenced issue(s): _n/a_

# Solution

Fix the linting errors one-by-one.

# Testing

```shell
npm run lint
```

### DIFF
--- a/packages/coinstac-images/.eslintrc.js
+++ b/packages/coinstac-images/.eslintrc.js
@@ -1,0 +1,5 @@
+module.exports = {
+  rules: {
+    'import/no-unresolved': ['off']
+  }
+}
--- a/packages/coinstac-ui/app/render/state/graphql/props.js
+++ b/packages/coinstac-ui/app/render/state/graphql/props.js
@@ -27,8 +27,9 @@ export const consortiaMembershipProp = (name) => {
             const consQuery = store.readQuery({ query: FETCH_ALL_CONSORTIA_QUERY });
             const index = consQuery.fetchAllConsortia.findIndex((cons) => {
               if (cons && data[name]) {
-                cons.id === data[name].id;
+                return cons.id === data[name].id;
               }
+              return false;
             });
             if (index > -1) {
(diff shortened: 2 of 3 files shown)


packages/coinstac-images/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
swashcap, line 1: Assuming all these subpackages are built as Docker images, so they won't have a _node_modules_ dir, causing the eslint-plugin-import to error. Disabling the rule seems like a sensible action.

packages/coinstac-ui/app/render/state/graphql/props.js
```diff
@@ -27,8 +27,9 @@ export const consortiaMembershipProp = (name) => {
             const consQuery = store.readQuery({ query: FETCH_ALL_CONSORTIA_QUERY });
             const index = consQuery.fetchAllConsortia.findIndex((cons) => {
               if (cons && data[name]) {
-                cons.id === data[name].id;
+                return cons.id === data[name].id;
```
swashcap, line 30: This changes program behavior! Could this be a bug where `index` is always `-1`?

packages/coinstac-images/.eslintrc.js
rssk: Yeah... I've been on the fence of how they should work, I think they're just fine as a free floating dir in packages that are then build as images. Maybe some update scripts to keep things consistent and make pubing images easier

packages/coinstac-ui/app/render/state/graphql/props.js
rssk: Great catch! I wonder... there are quite a few odd UI bugs

rssk (approved):

packages/coinstac-ui/app/render/state/graphql/props.js
swashcap: Thank the magic of ESLint 🔮

Issue #585, opened 2018-10-09
UI - multi users on consortia list page


rssk:
# Problem
If you add a new user to a consortia (and other actions too) the user list on the consortia list page shows several duplicates of users currently in the consortia

# Tasks
* correct the list view, wether its the list displaying incorrectly or the database data itself is wrong

rssk:
Closed by #586

PR #586, opened 2018-10-10
add or remove consUser instead of current user

abharathkr:
# Problem
referenced issue(s): #585 

# Solution
When a new user who is added to the consortium is made 'owner', current logged in user was being added or removed as owner and persisted into the database users table. Changed to add or remove the selected user as owner.


rssk (changes requested):
User list doesnt update when removed

abharathkr:
User list is now getting updated. Please review.

rssk (approved):

PR #587, opened 2018-10-11
Upgrade to React 16.x.x

swashcap:
# Problem

The UI uses an old version of React. Newer versions align with better testing tools and have improved performance.

referenced issue(s): _n/a_

# Solution

Upgrade the dependency, make the lifecycle method changes.

# Testing

I'm not sure how to fully test this. In practice, I've seen this upgrade subtly break components in ways that are difficult to determine without a full UI walkthrough. We should wait until E2E tests are working (see #584) before merging this.


swashcap:
Relevant blog post on some of the component lifecycle changes: https://reactjs.org/blog/2018/03/29/react-v-16-3.html

rssk:
Cool, I'll give this a thourough look through, I held off upgrading in the past for a reason I can't remember...... (of course)

swashcap:
@rssk Looks like the E2E test fixes are merged. Want to look at this one?

rssk:
Eduardo is trying to fix a drop down issue this caused, but besides that it isnt looking too bad

rssk (approved):

Issue #591, opened 2018-10-16
Observed Electron suddenly shutting down when starting Pipeline


pixelsaurus:
# Problem

Electron shutting down on Docker/Comp download

# Tasks

- Throw error in pull image possibly causing the issue
- Offending code possibly in app/main/index

PR #620, opened 2018-11-27
electron crash issues

abharathkr:
# Problem
referenced issue(s): #591 

# Solution

1. catch error in getImage and getStatus
2. pruneImages, removeImage already handle errors

### DIFF
--- a/packages/coinstac-ui/app/main/index.js
+++ b/packages/coinstac-ui/app/main/index.js
@@ -18,14 +18,6 @@ const ipcPromise = require('ipc-promise');
 const mock = require('../../test/e2e/mocks');
 const ipcFunctions = require('./utils/ipc-functions');
 
-// Require winston logger to log docker issues 
-const winston = require('winston');
-
-const winstonLogger = new winston.Logger({
-  level: 'verbose',
-  transports: [new winston.transports.Console()],
-});
-
 const { ipcMain } = electron;
 
 // if no env set prd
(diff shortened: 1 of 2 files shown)


packages/coinstac-ui/app/main/index.js
```diff
@@ -18,6 +18,14 @@ const ipcPromise = require('ipc-promise');
...
 const ipcFunctions = require('./utils/ipc-functions');
 
+// Require winston logger to log docker issues 
+const winston = require('winston');
+
+const winstonLogger = new winston.Logger({
+  level: 'verbose',
+  transports: [new winston.transports.Console()],
+});
+
```
rssk: Good setting up Winston, but there is already a logger instance setup that you should use. Look around line 75

packages/coinstac-ui/app/render/components/dashboard/dashboard.jsx
```diff
@@ -165,6 +165,13 @@ class Dashboard extends Component {
     });
 
     this.unsubscribeToUserMetadata = this.props.subscribeToUserMetaData(user.id);
+    
+    ipcRenderer.on('docker-error', (event, arg) => {
```
rssk, line 169: You'll also see there are ipc `unsubscribes` for the event handlers, you'll want one for this handler too

rssk (changes requested):

rssk (approved):

PR #638, opened 2019-01-16
File stream

### LINKED ISSUE #614: File transfer for computation communication
# Problem
The application cannot currently handle very large data transfers (> 256 MB) between computation steps between local and remote nodes.

# Tasks
Save communication payload from computation to file and transfer file between local and remote nodes.

rssk:
# Problem
referenced issue(s):
Closes #614 

Computations are limited by their out by the string length limit in JS, which is further compounded by the fact that remote site results are aggregated and serialized from the number of participating sites, meaning that limit can be reached rater quickly. Besides the fact that serialization and transmission is inefficient.

# Solution
*Don't do that

Instead, use the same socket stream that is in the docker node server to stream files from a new `transfer directory` directly into the other nodes input dir.
# Testing
* run the new file transfer image w/ sim


packages/coinstac-docker-manager/src/index.js
```diff
@@ -570,20 +570,24 @@ const removeImage = (imageId) => {
  * @return {Promise}              A promise that resovles on success
  */
 const stopService = (serviceId, serviceUserId) => {
-  if (services[serviceId].users.indexOf(serviceUserId) > -1) {
+  if (services[serviceId] && services[serviceId].users.indexOf(serviceUserId) > -1) {
```
rochaeb: For legibility sake I would put this `services[serviceId]` into a variable. Also I would do this on the beginning of the function `if (!services[serviceId]) return Promise.resolve()`, just to keep to the pattern of returning soon if you find an error.

packages/coinstac-pipeline/src/controller.js
```diff
@@ -250,13 +263,15 @@ module.exports = {
           const errCb = (err) => {
             setStateProp('state', 'error');
             controllerState.activeComputations[controllerState.computationIndex].stop()
-              .then(() => rej(err));
+              .then(() => rej(err))
+              .catch(err => rej(err));
```
rochaeb: The `err` variable is being overridden here. Just making sure you are aware.

packages/coinstac-pipeline/src/pipeline-manager.js
```diff
@@ -48,12 +53,23 @@ module.exports = {
...
     let socket;
     const remoteClients = {};
-    const missedCache = {};
+    // TODO: const missedCache = {};
 
     const waitingOnForRun = (runId) => {
       const waiters = [];
       activePipelines[runId].clients.forEach((client) => {
-        if (remoteClients[client][runId] && !remoteClients[client][runId].currentOutput) {
+        if ((remoteClients[client][runId]
```
rochaeb: putting this `remoteClients[client][runId]` into a variable would improve readability here and also probably make the code smaller.

packages/coinstac-pipeline/src/pipeline-manager.js
```diff
@@ -192,12 +255,64 @@ module.exports = {
         // TODO: step check?
         if (!data.error && activePipelines[data.runId]) {
           activePipelines[data.runId].state = 'recieved central node data';
-          activePipelines[data.runId].remote.resolve(data.output);
+          if (data.files) {
+            // we've already recieved the files
+            if (activePipelines[data.runId].files
+              && activePipelines[data.runId].files.recieved
```
rochaeb: Typo on `recieved`. Maybe you're not accessing the property you want?

rochaeb (approved):

packages/coinstac-pipeline/src/controller.js
rssk: I am, but maybe a bad choice of var names

packages/coinstac-docker-manager/src/index.js
rssk: 👍

packages/coinstac-pipeline/src/pipeline-manager.js
rssk: 👍

packages/coinstac-pipeline/src/pipeline-manager.js
rssk: I dont see the typo?

rssk: jk, errything is spelled wrong.....

PR #651, opened 2019-02-21
Let user select the home directory for the app

### LINKED ISSUE #646: UI: allow home dir to be configurable
# Problem
There are some issues with network mounted home dirs that dont allow cstac to properly function. As a solution, and a general user configuration, we should allow users to choose a 'home' cstac directory

# Tasks
* make the home directory configurable on startup
* * this presents some challenges, as the home dir will have to be configured before the users logs in (possibly) or before any such error from pipeline linking can happen. Perhaps in the system level dropdown you could change it? This change would also need to be persisted.

rochaeb:
# Problem
closes #646 

# Solution

# Testing

### DIFF
--- a/packages/coinstac-ui/app/render/state/ducks/auth.js
+++ b/packages/coinstac-ui/app/render/state/ducks/auth.js
@@ -3,6 +3,8 @@ import ipcPromise from 'ipc-promise';
 import { remote } from 'electron';
 import { applyAsyncLoading } from './loading';
 
+const CoinstacClientCore = require('coinstac-client-core');
+
 const apiServer = remote.getGlobal('config').get('apiServer');
 const API_URL = `${apiServer.protocol}//${apiServer.hostname}${apiServer.port ? `:${apiServer.port}` : ''}${apiServer.pathname}`;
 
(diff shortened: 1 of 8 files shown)


packages/coinstac-ui/app/render/state/ducks/auth.js
```diff
@@ -88,9 +101,13 @@ export const autoLogin = applyAsyncLoading(() => (dispatch) => {
...
 
-export const login = applyAsyncLoading(({ username, password, saveLogin }) => dispatch => axios.post(`${API_URL}/authenticate`, { username, password })
-  .then(({ data }) => initCoreAndSetToken({ username, password, saveLogin }, data, dispatch))
+export const login = applyAsyncLoading(({ username, password, saveLogin }) => (dispatch, getState) => axios.post(`${API_URL}/authenticate`, { username, password })
+  .then(({ data }) => {
+    const { auth: { appDirectory } } = getState();
+    return initCoreAndSetToken({ username, password, saveLogin }, data, appDirectory, dispatch);
+  })
   .catch((err) => {
+    caches.log('ER', err);
```
rssk: this intentional?

rssk (changes requested):

packages/coinstac-ui/app/render/state/ducks/auth.js
rochaeb: no :p

rssk (approved):

PR #652, opened 2019-02-21
Renamed some files for structure and organization

hvgazula:
# Problem
referenced issue(s):
something silly
# Solution
trying to mess with your heads
# Testing


rssk (changes requested):
Remove conflicts, not using a fork makes this easier in the future so I can do it....

rssk:
outdated

Issue #643, opened 2019-01-29
Stop pipelines


rssk:
# Problem
* pipelines can't be stopped

# Tasks
* create a function in pipeline manager that stops pipelines
* * the function should take a pipelineID and return promise for that pipeline stopping.
The function will need to emit a stopping event, removing all attached listeners to emitters, cleanup the active pipeline by ensuring its contents can be GC'd by either setting them `undefined` or using `delete`, and shutting down the associated docker boxes.

rssk:
Fixed in #657

PR #657, opened 2019-03-26
Stop pipeline

abharathkr:
# Problem
referenced feature(s): #643 

# Solution
Provided a "Stop Pipeline" button in consortia page for active pipeline
# Testing

### DIFF
--- a/packages/coinstac-client-core/src/index.js
+++ b/packages/coinstac-client-core/src/index.js
@@ -237,6 +237,20 @@ class CoinstacClient {
       });
   }
 
+  /**
+   * Stop pipeline
+   * @param {string} pipelineId The id of the pipeline running
+   * @param {string} runId The id of the pipeline run
+   * @param {string} userId The id of the user requesting pipeline to stop
+   */
+
+  stopPipeline( pipelineId, runId, userId ) {
+    const stopPipeline = this.pipelineManager.stopPipeline(pipelineId, runId);
+    return new Promise((res, rej) => {
+      return stopPipeline;
+    })
+  }
+
   unlinkFiles(runId) {
     const fullPath = path.join(this.appDirectory, this.clientId, runId);
 
--- a/packages/coinstac-client-core/src/index.js
+++ b/packages/coinstac-client-core/src/index.js
@@ -238,13 +238,14 @@ class CoinstacClient {
   }
 
   /**
-   * Stop pipeline
+   * Requests a pipeline stop by notifying the pipeline controller
+   * the pipeline will then stop via an error event and terminate as if an error
+   * were thrown
    * @param {string} pipelineId The id of the pipeline running
    * @param {string} runId The id of the pipeline run
    */
-  stopPipeline(pipelineId, runId) {
-    const result = this.pipelineManager.stopPipeline(pipelineId, runId);
-    return Promise.resolve(result);
+  requestPipelineStop(pipelineId, runId) {
+    this.pipelineManager.stopPipeline(pipelineId, runId);
   }
 
   unlinkFiles(runId) {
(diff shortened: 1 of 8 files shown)


packages/coinstac-client-core/src/index.js
```diff
@@ -237,6 +237,19 @@ class CoinstacClient {
...
 
+  /**
+   * Stop pipeline
+   * @param {string} pipelineId The id of the pipeline running
+   * @param {string} runId The id of the pipeline run
+   */
+
+  stopPipeline( pipelineId, runId ) {
+    const stopPipeline = this.pipelineManager.stopPipeline(pipelineId, runId);
+    return new Promise((res, rej) => {
```
rssk: Why create a new promise here? Also, this new promise doesnt resolve or reject anything, so you'll never actually get the async data to return right. `stopPipeline` should return a promise and thus you can just make this a one liner es6 arrow function

packages/coinstac-pipeline/src/controller.js
```diff
@@ -235,6 +239,10 @@ module.exports = {
           trampoline(() => {
             return queue.length
               ? function _cb(...args) {
+                console.log(controllerState);
```
rssk: ;-)

packages/coinstac-pipeline/src/controller.js
```diff
@@ -235,6 +239,10 @@ module.exports = {
...
+                if (controllerState.stopByUser == 'stop'){
+                  err({message: 'Pipeline run is stopped by user'});
```
rssk: It's good practice to actually create an error when error'ing out, as often upstream code checks for object type. 
`err(new Error('Pipeline run is stopped by user'))`

packages/coinstac-pipeline/src/pipeline-manager.js
```diff
@@ -541,6 +541,11 @@ module.exports = {
 
         return this.activePipelines[runId].stateEmitter;
       },
+      stopPipeline(pipelineId, runId) {
+        this.activePipelines[runId].pipeline.pipelineSteps.forEach( step => {
```
rssk: Couple things here, the pipeline-manager here is directly mussing w/ pipeline objects, I try to do this only when absolutely necessary. Also, there is only ever one currently running pipeline `step` and so you would only need to ever stop that one. 
What needs to happen is that the call to stop the currently running step needs to happen from manager to a pipeline function, and that function needs to call the currently active step's controller, thus stopping the pipeline.

rssk: Also also, docbloc your functions :-)

packages/coinstac-ui/app/main/index.js
```diff
@@ -238,6 +238,19 @@ loadConfig()
...
+    /**
+   * IPC Listener to stop pipeline
+   * @param {String} pipelineId The id of the pipeline currently running
+   * @param {String} runId The id of the pipeline run
+   * @return {Promise<String>} Status message
+   */
+
+    ipcMain.on('stop-pipeline', (event, { pipelineId, runId }) => {
+      return core.stopPipeline(pipelineId, runId)
+      .catch((err) => { console.log(`Pipeline could not be stopped because of the following error: ${err}`)});
```
rssk: this should probably be logged using the winston instance and then sent to the UI toast handler

rssk (changes requested):

rssk:
Thanks for cleaning this up, one more quick task:

Can you add a stop button to the home page where the pipeline running widget is?

rochaeb:
> Thanks for cleaning this up, one more quick task:
> 
> Can you add a stop button to the home page where the pipeline running widget is?

done

PR #666, opened 2019-04-09
Modifying Electron to add loading screen to alleviate user nervousness

pixelsaurus:
# Problem
referenced issue(s):

# Solution

# Testing

### DIFF
--- a/packages/coinstac-ui/app/main/utils/boot/configure-browser-window.js
+++ b/packages/coinstac-ui/app/main/utils/boot/configure-browser-window.js
@@ -16,12 +16,14 @@ const {
   BrowserWindow,
 } = electron;
 let mainWindow = null;
+let splashWindow = null;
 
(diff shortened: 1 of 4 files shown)


packages/coinstac-ui/app/main/utils/boot/configure-browser-window.js
```diff
@@ -30,13 +32,32 @@ function createWindow() {
...
     width: size.width,
     height: size.height,
+    show: false,
   });
 
   mainWindow.loadURL(`file://${renderIndexPath}`);
 
+  mainWindow.once('ready-to-show', () => {
+    mainWindow.show();
+    setTimeout(() => {
```
rssk, line 55: this works for now, but we should hide on an IPC from main once the libs are loaded

rssk (approved):

PR #668, opened 2019-04-18
Fix e2e tests

rochaeb:
# Problem
Fix e2e tests

# Solution
* Auto mapping was not working on e2e due to the speed of the automated test. I fixed the handling of async work to make it predictable.
* One element changed from `button` to `a`

# Testing


rssk:
Good work, there will probably be a fix for the auto map separately from Javier, but this works great for now

PR #699, opened 2019-07-31
Electron upgrade

rssk:
# Problem
Electron is old

# Solution
* Upgrade to v6.0.0, ups string max len to 1GB to give some headroom on computational output
* fix faulty file dialog
# Testing
* npm test
* run regression


hvgazula:
By the way, is this new change reflected in the simulator as well?

PR #717, opened 2019-09-04
Feature/698 pipeline update

### LINKED ISSUE #698: Saving an already created pipeline is broken
# Task

* When a pipeline has been saved, it should be able to be edited, instead it sometimes appears greyed out

* find out why its greyed out

* Correct functionality is:
  - the pipeline should always be editable
  - when the pipeline is set to active on the consortia pipeline page, editing the pipeline
     should un-map the current users who are mapped to the pipeline. This is already done if the 
     active pipeline is switched on the consortia pipeline page, look how this is done 

Perhaps relevant error:

Error:
collections.js:421 Uncaught (in promise) TypeError: col.associatedConsortia.indexOf is not a function
    at eval (collections.js:421)
    at Array.forEach (<anonymous>)
    at eval (collections.js:420)
    at Promise.js:801
    at callListener (Promise.js:469)
    at endMicroTickScope (Promise.js:555)
    at IDBTransaction.<anonymous> (Promise.js:624)

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #698

### DIFF
--- a/packages/coinstac-ui/app/render/utils/helpers.js
+++ b/packages/coinstac-ui/app/render/utils/helpers.js
@@ -1,0 +1,5 @@
+import { indexOf } from 'lodash';
+
+export const isPipelineOwner = (permissions, owningConsortium) => {
+  return indexOf(permissions.consortia[owningConsortium], 'owner') !== -1
+}
--- a/packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
+++ b/packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
@@ -6,7 +6,7 @@ import HTML5Backend from 'react-dnd-html5-backend';
 import PropTypes from 'prop-types';
 import { graphql, withApollo } from 'react-apollo';
 import shortid from 'shortid';
-import { isEqual, isEmpty } from 'lodash';
+import { isEqual, isEmpty, get } from 'lodash';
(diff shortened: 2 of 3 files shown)


packages/coinstac-ui/app/render/utils/helpers.js
```diff
@@ -0,0 +1,5 @@
+import { indexOf } from 'lodash';
+
+export const isPipelineOwner = (permissions, owningConsortium) => {
+  return indexOf(permissions.consortia[owningConsortium], 'owner') !== -1
```
rochaeb, line 4: The old if's were looking into the `write` property and you changed to the `owner` property. Better take a look at this and make sure it's right.

rochaeb (changes requested):

packages/coinstac-ui/app/render/utils/helpers.js
kalitopix: @rochaeb , I already checked it and it works.

Currently, permissions.consortia[owningConsortium] returns array like this - ['member', 'owner'] and using *.write does not make sense.

Please check this on your side with logging.

rochaeb: Fixed with PR #731

rochaeb (approved):

PR #723, opened 2019-09-07
Feature/722 active pipeline delete

### LINKED ISSUE #722: UI: deleting an active pipeline causes an error
# Task
* create any pipeline, the `decentralized test` is simple to fill out
* set active on a consortia
* delete the pipeline
* observe error

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #722

### DIFF
--- a/packages/coinstac-api-server/src/data/resolvers.js
+++ b/packages/coinstac-api-server/src/data/resolvers.js
@@ -450,7 +450,7 @@ const resolvers = {
               .delete({ returnChanges: true })
               .run(connection),
               rethink.table('consortia').filter({ activePipelineId: args.pipelineId })
-              .update({ activePipelineId: '' })
+              .replace(rethink.row.without('activePipelineId'))
               .run(connection)
             ])
           }
--- a/packages/coinstac-api-server/src/data/resolvers.js
+++ b/packages/coinstac-api-server/src/data/resolvers.js
@@ -444,12 +444,18 @@ const resolvers = {
           ) {
             return Boom.forbidden('Action not permitted');
           } else {
-            return rethink.table('pipelines').get(args.pipelineId)
+            return new Promise.all([
+              connection,
+              rethink.table('pipelines').get(args.pipelineId)
               .delete({ returnChanges: true })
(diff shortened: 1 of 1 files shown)


packages/coinstac-api-server/src/data/resolvers.js
```diff
@@ -444,12 +444,18 @@ const resolvers = {
...
             return Boom.forbidden('Action not permitted');
           } else {
-            return rethink.table('pipelines').get(args.pipelineId)
+            return new Promise.all([
+              connection,
+              rethink.table('pipelines').get(args.pipelineId)
               .delete({ returnChanges: true })
-              .run(connection).then(res => connection.close().then(() => res))
+              .run(connection),
+              rethink.table('consortia').filter({ pipelineId: args.pipelineId })
```
rochaeb: Consortia table does not have a field called `pipelineId`. There's actually one called `activePipelineId`. Also there's a field called `pipelines` which is an array of id's of pipelines, which may also need to be updated.

Also, I don't know if updating the field to empty string will work. Maybe you should be using null here.

rochaeb (changes requested):

packages/coinstac-api-server/src/data/resolvers.js
kalitopix: @rochaeb 
I updated field name to `activePipelineId`.

By the way,  I cannot see `pipelines` field in Consortia table.
This is current Consortia table structure.
`
{
  "activeComputationInputs": [],
  "activePipelineId": "",
  "description": "ACON",
  "id": "0fd110df-4528-466b-bbf1-b1cf856daae9",
  "members": [
    "test1"
  ],
  "name": "ACon",
  "owners": [
    "test1"
  ],
  "tags": []
}`

kalitopix: Updating the `activePipelineId` field to empty string works (checked on my side)
But I updated codebase to remove `activePipelineId` key when the active pipeline is deleted from consortia to get rid of your worries, @rochaeb

rochaeb (approved):

PR #725, opened 2019-09-08
Made changes to FSL and VBM related schemas and docker images

hvgazula:
# Problem
referenced issue(s):
Inconsistent names for computations and their corresponding docker images
# Solution
Corrected all schemas and their corresponding docker images (that needs to be addressed in dockerhub)

# Testing


rssk (approved):
thanks!

PR #731, opened 2019-09-12
Remove roles table

### LINKED ISSUE #728: Simplify user permissions handling
# Problem
Currently, user permissions are a little too complex because they join 2 tables *users* and *roles*.

# Tasks
We can make it simpler by removing the *roles* table as it does not have a purpose at the moment.

rochaeb:
# Problem
closes #728 

# Solution

# Testing


rssk (approved):

rssk:
Looks great!

PR #732, opened 2019-09-13
added a mapping check to consortia

### LINKED ISSUE #730: UI: add a mapping check
Screen Shot 2019-09-11 at 12 03 37 PM
# Task

It's not easy to tell when users are ready/mapped, putting a checkmark on the avatar picture seems like a good solution, see attached photo.

* To do this you'll need to add a column in the `consortia` db table, something like `mappedForRun`
* the users will update that table and add if they're mapped from the `maps` page
* unmap actions should also remove the user from that list, check out how unmap works on active pipeline change

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #730


rssk:
Looks good! just add the unmap portion too

packages/coinstac-api-server/src/data/shared-fields.js
```diff
@@ -10,6 +10,7 @@ const schemaFields = {
     members: [ID]
     pipelines: [ID]
     results: [ID]
+    mappedForRun: [String]
```
rochaeb: should be array of ID here

packages/coinstac-ui/app/render/components/maps/maps-edit.jsx
```diff
@@ -266,11 +268,20 @@ class MapsEdit extends Component {
...
+
+    if (mappedForRun.indexOf(currentUserId) === -1) {
+      mappedForRun = [...mappedForRun, currentUserId]
+    }
 
     this.props.getCollectionFiles(cons.id)
       .then((filesArray) => {
         this.setState({isMapped: true});
 
+        this.props.saveConsortium({ id: cons.id, name: cons.name, description: cons.description, mappedForRun });
```
rochaeb: no need to send name and description here

kalitopix: @rochaeb , name and description are required fields for consortia

rochaeb: So I'd say create another mutation for this, to improve readability. Using this mutation as is can be misleading as you're sending unnecessary stuff...

rochaeb:
it's looking good, but I think that unmapping only when another pipeline is set is not enough... you'd definitely need to unmap if a user gets kicked out of a consortia or chooses to leave the consortia. And also maybe on user logout? @rssk, what do you think?

rssk:
I think we talked about this a bit over slack, but I think all the unmapping should be moved to one call, otherwise this is going to get out of hand. I think for now, I'd like that to do both local and remote map/unmap, it really isnt any more overhead and can always be changed later based on our local store reqs. What do you think @rochaeb @kalitopix

rssk (changes requested):
Add a sane way to unmap local an remote

packages/coinstac-ui/app/render/components/maps/maps-edit.jsx
kalitopix: @rochaeb , do you think creating another mutation is better than sending unnecessary info?
what about you, @rssk ?

kalitopix:
@rssk , @rochaeb 
I created a mutation that updates the mapping status of consortium.

Do we need to unmap user if he gets kicked out of a consortia or chooses to leave the consortia or maybe on logout as @rochaeb suggested?

packages/coinstac-ui/app/render/components/maps/maps-edit.jsx
rochaeb: yeah, I think so

kalitopix: @rochaeb , it is already done

rssk (approved):
revisions look good!

PR #737, opened 2019-09-30
Fix file dialog mock

rochaeb:
# Problem
Fix file dialog mock usage on e2e tests.

# Solution

# Testing


kalitopix (approved):
Awesome, @rochaeb

PR #746, opened 2019-10-18
Remove splash screen from separate window

rochaeb:
# Problem
Make the splash screen part of the same window as the app

# Solution

# Testing


rssk (changes requested):
This is super awesome, thanks for getting this lookin better! Fix the merge, and also there's an thrown error on logout that's causing an bad error message on the login page, it might of been there before? but I'm not sure

rssk (commented):
gr8 :0)

PR #756, opened 2019-10-30
UI Mapping - Removing extra columns from metaFile manually and via autoMap

pixelsaurus:
# Problem
Re: Ross: "Basically the mapping mapped all the columns, regardless of what you x’d out". This removes unnecessary columns from metaFile copy stored in Coinstac. This should resolve last week's 'Mapping UI Issues + PDF Gen Fixes' PR.


rssk (approved):
Looks like this is looking good now

PR #757, opened 2019-11-04
Do not redirect to login page on bad auth token

rochaeb:
# Problem
referenced issue(s):

# Solution

# Testing


rssk (approved):
looks great!

PR #770, opened 2019-11-12
Feature/693 result button

### LINKED ISSUE #693: Get Results button
# Problem
Users want to get to their results after a pipeline run, but they don't know where they are stored locally on their machines.

# Tasks
Put a button on the pipeline results page that takes the user to their data. Not sure about the best way to do this. Could open a finder/explorer window, could copy the data to the downloads folder, etc.

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #693

### DIFF
--- a/packages/coinstac-ui/app/render/components/common/run-item.jsx
+++ b/packages/coinstac-ui/app/render/components/common/run-item.jsx
@@ -1,6 +1,9 @@
 import React, { Component } from 'react';
 import TimeStamp from 'react-timestamp';
 import { Link } from 'react-router';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+import { shell } from 'electron';
 import Paper from '@material-ui/core/Paper';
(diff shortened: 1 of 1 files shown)


rochaeb (approved):

packages/coinstac-ui/app/render/components/common/run-item.jsx
```diff
@@ -116,198 +122,221 @@ function getStateWell(runObject, stateName, stateKey, classes) {
...
-          </Typography>
-        )
-      }
-      {
-        runObject.endDate
-        && (
+export class RunItem extends Component {
+  handleOpenResult = () => {
+    const { runObject, appDirectory, user } = this.props;
+    const resultDir = `${appDirectory}/output/${user.id}/${runObject.id}`
```
rssk: probably should be `path.join`'d here

rssk (changes requested):

PR #791, opened 2019-12-13
Integrating GICA computation and adding ability to destroy map from map-list view

pixelsaurus:
# Problem and Solution
Same as above


rssk (approved):
Looks good!

PR #795, opened 2020-01-10
Feature/pipeline timeout

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing


rssk (changes requested):
This is looking good! The next step is to wire up the `timeout` field to GQL and have it saved and persisted along with all the other pipeline data.

rssk (approved):

PR #797, opened 2020-01-14
Feature/796 password reset

### LINKED ISSUE #796: Password reset
# Problem
Need password reset
# Tasks
Allow users to update their password.
Requires current password to set new one.

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #796


packages/coinstac-api-server/src/auth-helpers.js
```diff
@@ -63,6 +77,32 @@ const helperFunctions = {
...
+      subject: 'Password Reset Request',
+      html: `We received your password reset request. <br/>
+        Please use this token for password reset. <br/>
+        Token: <strong>${resetToken}</strong>`,
+    };
+
+    const db = database.getDbInstance();
+
+    await sgMail.send(msg);
+    await db.collection('users').findOneAndUpdate({ email }, {
```
rochaeb: Since you do not want to get the user after the operation you should use updateOne. Also, you don't need to wait for the operation... just return it.

packages/coinstac-api-server/src/auth-helpers.js
```diff
@@ -205,6 +245,64 @@ const helperFunctions = {
       res(Boom.unauthorized('Incorrect username or password.'));
     }
   },
+  /**
+   * Confirms that submitted email is valid
+   * @param {object} req request
+   * @param {object} res response
+   * @return {object} The requested object
+   */
+  async validateEmail(req, res) {
```
rochaeb: I'd suggest here to not create this function. You could use the already existing `validateUniqueEmail` function, which does the same thing. Maybe rename it if you want.

packages/coinstac-api-server/src/auth-helpers.js
```diff
@@ -63,6 +77,32 @@ const helperFunctions = {
...
+        Please use this token for password reset. <br/>
+        Token: <strong>${resetToken}</strong>`,
+    };
+
+    const db = database.getDbInstance();
+
+    await sgMail.send(msg);
+    await db.collection('users').findOneAndUpdate({ email }, {
+      $set: {
+        passwordResetToken: resetToken,
```
rochaeb, line 86: what about adding a timeout to the token? Something like a couple hours or so...

packages/coinstac-api-server/src/auth-helpers.js
```diff
@@ -205,6 +245,64 @@ const helperFunctions = {
...
+   * Reset password
+   * @param {object} password token for resetting password
+   * @param {object} password new password
+   * @return {object}
+   */
+  async resetPassword(token, password) {
+    const db = database.getDbInstance();
+
+    const newPassword = await helperFunctions.hashPassword(password);
+    await db.collection('users').findOneAndUpdate({
```
rochaeb: Same thing here. You're not expecting the return from this function, so use `updateOne` instead.

packages/coinstac-api-server/src/auth-helpers.js
rssk: 👍

kalitopix: token has expiresIn itself

rssk (approved):

PR #809, opened 2020-01-30
Converted sync file code async

### LINKED ISSUE #355: Client-Core: make sync file code async
# Problem
The recursive `getSubPathsAndGroupExtension ` function uses a sync file method
# Solution
Make it async, note that excessively large path structures will cause an overflow (1000+ nested)

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #355

### DIFF
--- a/packages/coinstac-client-core/src/index.js
+++ b/packages/coinstac-client-core/src/index.js
@@ -193,7 +193,8 @@ class CoinstacClient {
       const stats = await fs.statAsync(p);
 
       if (stats.isDirectory()) {
-        const paths = [...fs.readdirSync(p).filter(item => !(/(^|\/)\.[^\/\.]/g).test(item))] // eslint-disable-line no-useless-escape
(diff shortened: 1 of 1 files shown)


packages/coinstac-client-core/src/index.js
```diff
@@ -190,12 +190,13 @@ class CoinstacClient {
         p = group.parentDir.concat(`/${p}`);
       }
 
-      const stats = fs.statSync(p);
+      const stats = await fs.statAsync(p);
 
       if (stats.isDirectory()) {
+        const paths = [...fs.readdirSync(p).filter(item => !(/(^|\/)\.[^\/\.]/g).test(item))] // eslint-disable-line no-useless-escape
```
rssk: `fs.readdirSync(p)` should be converted too :-)

rssk (changes requested):

PR #812, opened 2020-02-03
Fix/364 unique pipeline consortium

### LINKED ISSUE #364: Unique Pipeline and Consortia Names
# Problem
Pipeline and Consortia names should be unique.
# Solution
- Check before saving either via their respective `save` GraphQL calls or some sort of `onBlur` action called when the user leaves the name field.
- Display a Bootstrap alert to users when a duplicate name is used.

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #364


packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
```diff
@@ -437,12 +437,11 @@ class Pipeline extends Component {
       });
 
       notifySuccess({ message: 'Pipeline Saved.' });
-    }).catch((error) => {
-      notifyError({ message: error.message });
+    }).catch(({ graphQLErrors }) => {
+      console.log(graphQLErrors);
```
rssk: this log intended?

rssk (changes requested):

packages/coinstac-ui/app/render/components/pipelines/pipeline.jsx
kalitopix: yes. want to remove?

rssk (approved):

PR #813, opened 2020-02-04
add easier dev tools for starting cstac

rssk:
# Problem
Dev setup is a pain
# Solution
Fix by adding a better start option, maybe a vscode script in the future?


pixelsaurus:
I have a shell script that I use to get a dev Coinstac running. Would that work?

PR #821, opened 2020-02-25
Fix test

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

### DIFF
--- a/packages/coinstac-api-server/src/data/schema.js
+++ b/packages/coinstac-api-server/src/data/schema.js
@@ -164,7 +164,7 @@ const typeDefs = `
     leaveConsortium(consortiumId: ID!): Consortium
     removeComputation(computationId: ID): Computation
     removeUserRole(userId: ID!, table: String!, doc: String!, role: String!): User
-    saveActivePipeline(consortiumId: ID, activePipelineId: ID): String
+    saveActivePipeline(consortiumId: ID, activePipelineId: ID): JSON
     saveConsortium(consortium: ConsortiumInput!): Consortium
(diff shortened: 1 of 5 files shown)


packages/coinstac-ui/app/render/state/ducks/docker.js
```diff
@@ -38,6 +38,8 @@ export const removeImage = applyAsyncLoading((compId, imgName, imgId) => dispatc
   }));
 
 export const updateDockerOutput = (output => (dispatch) => {
+  console.log('------------------------')
+  console.log(output)
```
rssk: remove?

packages/coinstac-api-server/src/data/schema.js
```diff
@@ -164,7 +164,7 @@ const typeDefs = `
     leaveConsortium(consortiumId: ID!): Consortium
     removeComputation(computationId: ID): Computation
     removeUserRole(userId: ID!, table: String!, doc: String!, role: String!): User
-    saveActivePipeline(consortiumId: ID, activePipelineId: ID): String
+    saveActivePipeline(consortiumId: ID, activePipelineId: ID): JSON
```
rssk, line 167: whats up with this change?

rssk (changes requested):

packages/coinstac-api-server/src/data/schema.js
kalitopix: check **saveActivePipeline** in resolvers.
It returns updated result and it is JSON not string.

PR #826, opened 2020-03-02
Fix some start pipeline bugs

rochaeb:
# Problem
- In some scenarios it was possible to start a pipeline without a proper data mapping, which would lead to errors lately.
- UI would not allow the user to start a pipeline that don't need data map

# Solution

# Testing


packages/coinstac-ui/app/main/index.js
```diff
@@ -345,6 +321,37 @@ loadConfig()
...
+
+        const run = {
+          ...pipelineRun,
+          pipelineSnapshot: {
+            ...pipelineRun.pipelineSnapshot,
+            steps,
+          },
+        };
+
+        startPipelineRun(run, filesArray, consortium);
```
rssk: this is async, should either be awaited or caught via prom

rssk (changes requested):

rssk (approved):

PR #857, opened 2020-04-09
Feature/855 upload compspecs

### LINKED ISSUE #855: Feature: add the ability for authors to upload compspecs
# Problem
* there is no way for authors to personally upload compspecs for updated computations
# Tasks
* make a gql endpoint for uploading computation specs to the database, the computation table already exists and you can see examples in the `test-data` in `coinstac-api-server`. The endpoint takes a json document such as this https://github.com/trendscenter/coinstac-dpsvm/blob/master/compspec.json
* create two new user roles `author` and `admin` (author might exist? I'd check)
* only allow those with the author role to upload new compspecs or overwrite a compspec they uploaded ( they cannot overwrite a compspec not uploaded by them and should be given a clear error message)
* create an option `--upload` with shortname `-u` in `coinstac-simulator` that asks for username and password, and then uploads the `compspec.json` in the directory its called in (ie: https://github.com/trendscenter/coinstac-dpsvm/blob/master/compspec.json) to the gql endpoint
* create an Admin page in dashboard that allows you to take a username and give them the `author` role. This page should only be visible to users with the `admin` role

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #855


rssk (changes requested):
I can't seem to get changing user permissions to work on the permissions page

rssk (approved):

PR #863, opened 2020-04-16
Fix/858 pipeline server down

### LINKED ISSUE #858: UI: pipeline should error if pipeline server is down
# Problem
* pipeline hangs if the pipeline server is not reachable, it should just error out instead

# Tasks
* error out in UI and return a meaningful message

kalitopix:
# Problem
referenced issue(s):

# Solution

# Testing

closes #858


rssk (changes requested):
This looks good, however I can't seem to get it to stop the local pipeline but it does show an error

rssk:
Closing do to nonsensical diff, will recreate

Issue #866, opened 2020-04-21
UI - Bundle mapping not working for ENIGMA


pixelsaurus:
**Problem**

For whatever reason a refactor in the function getSubPathsAndGroupExtension in coinstac-client-core/src/index.js broke mapping for and bundled set of files.  I noticed this while working on dsne-ms.

**Solution**

Revert this function to the last working version.

rssk:
Resolved

PR #867, opened 2020-04-21
Reverting function "getSubPathsAndGroupExtension" to working version

pixelsaurus:
# Problem
referenced issue(s): #866 

# Solution

Reverted to a working version of the function

# Testing

Create an Enigma pipeline and map to it


rssk:
Fixed in #868

PR #909, opened 2020-07-08
Fix mapped status and add online status check

rochaeb:
# Problem
referenced issue(s):

# Solution

# Testing


rssk (changes requested):
the user status sub is returning all users statuses, lets filter it by the current users consortium (as in any consortia they may be joined, so possibly mutiple)

rssk (approved):

PR #912, opened 2020-07-15
Pipeline: fix major memory issues

rssk:
# Problem
* the pipeline has a leak

# Solution
* Large portions of the pipeline had to be modified, namely the way output was passed around, using a new centralized store to minimize reference pollution. 
* trace and clarify seem to hold on to large portions of references causing more issues than they're worth

# Testing
* run drne and watch memory usage


packages/coinstac-docker-manager/package.json
```diff
@@ -9,13 +9,16 @@
...
   "license": "ISC",
   "dependencies": {
+    "clarify": "^2.1.0",
     "coinstac-socket.io-stream": "^0.9.1",
     "dockerode": "^2.5.5",
     "lodash": "^4.17.10",
     "portscanner": "^2.2.0",
     "request-stream": "^1.2.2",
-    "socket.io-client": "^2.1.1",
+    "socket.io-client": "^2.3.0",
```
rochaeb: is this being used?

packages/coinstac-simulator/src/index.js
```diff
@@ -16,16 +17,18 @@ const mosca = require('mosca');
...
-
-  server.on('clientConnected', (client) => {
-    console.log('Mosca client connected', client.id); // eslint-disable-line no-console
-  });
-
-  return new Promise((resolve) => {
-    server.on('ready', resolve);
-  })
-    .then(async () => {
+  // return new Promise((resolve, reject) => {
```
rochaeb: nice comment

rssk: :P

packages/coinstac-docker-manager/package.json
rssk: nah, there is a couple more things that need to be pruned too

PR #925, opened 2020-07-23
Fix default values on pipeline creation

### LINKED ISSUE #905: fMRI preprocess can't map data
# Problem
When I click Map Data on the consortium after creating the fMRI preprocess pipeline, the screen goes blank, and I have to restart the application. Reload doesn't help. I don't see any meaningful errors in the COINSTAC logs, but I see these errors in the Dev Tools console.
Screen Shot 2020-07-01 at 16 20 04

# Tasks
Fix the bug.

rochaeb:
# Problem
closes #905 

# Solution

# Testing


rssk (changes requested):
Conflicts

PR #1006, opened 2021-01-29
Fix/1002 pipeline username updates

### LINKED ISSUE #1002: UI: usernames no longer displayed on pipeline updates and errors
## Expected Behavior

Usernames should be displayed on pipeline run progress updates eg: `Waiting on user1, user3....`, but they are now database IDs, the same goes for when the pipeline errors out eg: `Error from User2.....`

## Possible Solution

Change the pipeline run updates to include the usernames, this may necessitate changing the update structure to have the usersids in a obj. For errors.... you'll probably have to do a regex find and replace

## Steps to Reproduce
Run a normal run and cause a user(s) to be late, you could also just look at the code in pipeline-manager to see how it generates that list. Error is simpler, just cause an error from a local node:-)

## Context (Environment)
UI/pipeline

kalitopix:
- [x] Passes e2e testing

Show usernames correctly for waiting list and errored runs.

referenced issue(s):
closes #1002 

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
Yes, updated startPipeline to use object based clients as a parameter.


rssk (approved):
Looks great! thanks for taking care of this

PR #1038, opened 2021-04-02
Feature/1035 result file browser

### LINKED ISSUE #1035: UI: File browser for results
## Expected Behavior
There isn't an easy and clear way to browse files currently saved under results.

A new file `tab` in the results page should provide this

## Current Behavior
There is currently on a button to browse a specific run's directory

## Possible Implementation
* create a tab on the results page to delineate the file browsing mode
* create a UI area resembling a file tree like so:
```
| Consortia Name
    | Pipeline Name - RunID - RunDate
        | -> file1
        | -> file2 etc
    | Another Pipeline Name - RunID - RunDate
        | -> file1
        | -> file2 etc
| Another Consortia Name
    | Yet Another Pipeline Name2 - RunID - RunDate
        | -> file1
        | -> file2 etc
```
* in order to facilitate this you'll have to programmatically create this structure in the coinstac home directory for that user (check in the ui/config.js file for how that's set). This should be an artificial directory structure created from the output directory, as the output directory uses non human readable dir names for pipeline to function correctly, basically just create the `Consortia/Run` dirs then symlink the files over. Maybe do this on startup? Or maybe on that result tabs load, should be quick and only needs to be done if there is new stuff.

kalitopix:
- [x] Passes lint

Feature

referenced issue(s):
closes #1035 

Added file browser for consortia results


packages/coinstac-ui/app/render/components/results/file-viewer.jsx
```diff
@@ -0,0 +1,178 @@
...
+        appDirectory: auth.appDirectory,
+      });
+    }
+  }, [auth, initialFileTree]);
+
+  useEffect(() => {
+    ipcRenderer.on('prepare-consortia-files', (event, nodes) => {
+      setFileTree(buildTree(nodes));
+    });
+  }, []);
```
rssk: A couple things here, I think we really only need to build the file tree on page load, not on each render.

Also `useMemo` and `useEffect` don't deep compare your change values, so they'll always run I think

packages/coinstac-ui/app/main/index.js
```diff
@@ -124,6 +136,65 @@ loadConfig()
...
+      }
+
+      const res = [];
+
+      fileTree.forEach((consortium) => {
+        const consortiumDirectory = path.join(userRunDirectory, consortium.name);
+
+        if (!fs.existsSync(consortiumDirectory)) {
+          fs.mkdirSync(consortiumDirectory);
+        }
```
rssk, line 157: async:-)

packages/coinstac-ui/app/main/index.js
```diff
@@ -124,6 +136,65 @@ loadConfig()
...
+
+        consortium.runs.forEach((run) => {
+          const runDirectory = path.join(consortiumDirectory, `${run.pipelineName} - ${run.id}`);
```
rssk: Actually you know what, can we just append the run end date?

packages/coinstac-ui/app/main/index.js
```diff
@@ -124,6 +136,65 @@ loadConfig()
...
+            fs.mkdirSync(runDirectory, { recursive: true });
+          }
+
+          const outputDirectory = path.join(appDirectory, 'output', userId, run.id);
+
+          if (fs.existsSync(outputDirectory)) {
+            const allFiles = getAllFilesInDirectory(outputDirectory);
+
+            allFiles.forEach((file) => {
+              const relativePath = file.replace(outputDirectory, '');
```
rssk: `path.basename()`

rssk: actually, you want this path relative to the `runId` dir so you would use `path.relative(from, to)`

packages/coinstac-ui/app/main/index.js
```diff
@@ -124,6 +136,65 @@ loadConfig()
...
+              }
+
+              if (!fs.existsSync(symlinkPath)) {
+                fs.symlinkSync(file, symlinkPath);
+              }
+
+              let createdSymlinkPath = symlinkPath.replace(appDirectory, '');
+              if (createdSymlinkPath[0] === '/') {
+                createdSymlinkPath = createdSymlinkPath.slice(1);
+              }
```
rssk: `path.relative` should make this a one liner and platform agnostic:-)

packages/coinstac-ui/app/main/index.js
kalitopix: any reason to use async?

kalitopix: `fs.exists` is deprecated and `fsPromise` does not have `exists`.
we have to use `fs.existsSync` and if so, I would like to use `fs.mkdirSync` for consistency.

btw, I don't see any needs to use async here. @rssk

PR #1188, opened 2022-02-04
Feature validate spec

### LINKED ISSUE #362: Simulator: checks for specs using joi
# Problem
It's easy to get weird errors w/ a bad spec, reject bad specs w/ info why theyre bad

### LINKED ISSUE #1184: Validate spec on save pipeline
**Is your feature request related to a problem? Please describe.**
Pipelines are able to be saved even when computation specs have not been filled out correctly.
For instance:
in the UI for Ridge Regression (Singleshot) - FreeSurfer Volumes,
If no covariates are specified, the pipeline can still be saved. 

**Describe the solution you'd like**
Using a validation library, the ?inputspec? can be compared to the compspec to determine if it is valid. If a spec is determined to be invalid, the user should be prompted to fix it before saving the pipeline.

dylanmartin:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Passes e2e testing
- [x] Passes lint
- [x] Docs have been added / updated if necessary (for bug fixes / features)

Adds validation to inputMap when saving a pipeline.

closes _issue #_1184

Adds validation to inputMap when saving a pipeline.


rssk:
So I think a good goal here would be to get SSR validating fully. What that'll mean is that on pipeline save the owner fields need to be filled out to a minimum, this might just mean checking if the obj exists rather than the entire semantic fulfillment, and on mapping which will mean member fields have also been filled out to the same standard

dylanmartin:
This validates on the input spec when it comes from the UI when saving a pipeline. Is the validation that happens here identical to what should happen when uploading a spec in simulator? is inputMap in the same form here as in other places? Aside from the UI, where should the validator be used?

PR #1237, opened 2022-03-29
Allow selecting subset of users for consortium runs

rochaeb:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Passes e2e testing
- [x] Passes lint
- [ ] Docs have been added / updated if necessary (for bug fixes / features)

feature

All consortium members must take part in runs

Allow selecting a subset of users for future runs in consortium

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
no

* **Other information**:


packages/coinstac-api-server/src/data/controllers/headlessClients.js
```diff
@@ -43,6 +43,19 @@ async function _removeOwnerPermissionFromUsers(headlessClientId, userIds) {
   eventEmitter.emit(USER_CHANGED, users);
 }
 
+async function _removeActiveConsortiaMembers(headlessClientId) {
```
rssk: this function seems to be named a little oddly, unless im misunderstanding, it seems to remove one member (the headless one) from multiple consortium

packages/coinstac-api-server/src/data/resolvers.js
```diff
@@ -927,16 +942,47 @@ const resolvers = {
...
+          Object.keys(oldPipeline.headlessMembers).forEach((headlessMemberId) => {
+            oldPipelineUpdateObj.$unset[`activeMembers.${headlessMemberId}`] = '';
+          });
+
+          await db.collection('consortia').updateOne(
+            { _id: ObjectID(args.consortiumId) },
+            oldPipelineUpdateObj
+          );
+        }
+      }
```
rssk, line 965: i might be misunderstanding this too, but why not just clear active members entirely, rather than piecemeal removing the headless ones? Dont the owners choose new active members with a new active pipeline? If they are being preserved between pipelines, then this makes sense

packages/coinstac-api-server/src/data/resolvers.js
```diff
@@ -1072,6 +1154,48 @@ const resolvers = {
...
+          updateObj
+        );
+
+        consortiumChanged = true;
+      }
+
+      if (consortiumChanged) {
+        const consortium = await db.collection('consortia').findOne({ activePipelineId: args.pipeline.id });
+        eventEmitter.emit(CONSORTIUM_CHANGED, consortium);
+      }
```
rssk: why is save pipeline setting an active pipeline id?

packages/coinstac-api-server/src/database/migrations/scripts/2022-03-28-add-active-members-to-consortium.js
```diff
@@ -0,0 +1,24 @@
...
+
+  db.collection('consortia').updateMany({}, [
+    { $set: { activeMembers: "$members" } }
+  ]);
+}
+
+async function down() {
+  const db = database.getDbInstance();
+
+  // Downgrade code goes here
```
rssk: shouldnt this unset the obj?

packages/coinstac-ui/app/render/components/consortia/consortium-about.jsx
```diff
@@ -124,7 +124,7 @@ class ConsortiumAbout extends Component {
     });
 
     removeUserRole(user.id, 'consortia', consortium.id, 'member', 'data').then(({ data }) => {
-      updateConsortium({ param: 'members', value: omit(consortium.owners, [data.removeUserRole.id]) });
+      updateConsortium({ param: 'members', value: omit(consortium.members, [data.removeUserRole.id]) });
```
rssk, line 127: was this a bug?

rssk (changes requested):

packages/coinstac-ui/app/render/components/consortia/consortium-about.jsx
rochaeb: yeah

packages/coinstac-api-server/src/data/controllers/headlessClients.js
rochaeb: yeah, it removes the headless client that is being deleted from all consortia where it's a active member. I'll try to come up with a better name.

packages/coinstac-api-server/src/data/resolvers.js
rochaeb: The way I did it selecting the active members is not a required step, so whenever the user changes the active pipeline I try to keep the selection for users and just set the headless members as a default.

packages/coinstac-api-server/src/data/resolvers.js
rochaeb: It's using the active pipeline as a filter. The idea here is to whenever you edit a pipeline that is active in any consortium, the active members for this consortium should be updated by removing the old headless members and adding the new ones.

packages/coinstac-api-server/src/database/migrations/scripts/2022-03-28-add-active-members-to-consortium.js
rochaeb: yeah... It's just that we don't have any use for downgrade right now, but I'll add the code for it.

packages/coinstac-api-server/src/data/controllers/headlessClients.js
rssk: 👍

packages/coinstac-api-server/src/data/resolvers.js
rssk: I think that sounds fine to me

packages/coinstac-api-server/src/data/resolvers.js
rssk: ah, that makes sense

packages/coinstac-api-server/src/database/migrations/scripts/2022-03-28-add-active-members-to-consortium.js
rssk: nbd

Issue #1350, opened 2022-05-31
Pipeline: fix end file transfer 


rssk:
## Task Description
When on the final round trip, as in the remote as declared 'success', the files from the remote to the outer nodes are not tied to any  promise/await and as such are unable to be guaranteed to be done at the end of the pipeline promise

## Task Notes
* Tie file transfer into the pipeline promise, allowing the UI/etc to know that files are done
* look here https://github.com/trendscenter/coinstac/blob/master/packages/coinstac-pipeline/src/setup-outer.js#L404

## Steps to Reproduce
Run any pipeline (modify large file test? in `packages/coinstac-images`) with large files and observe the file system vs end difference

rssk:
Note: looking through the code this _might_ be fixed, but I'm unclear, this is easiest to get breaking using a dev/prod build as files have to traverse the network

dylanmartin:
The code does look like it handles this case but it's worth checking.

Can I get more explicit details on reproducing the error? 
* Which computation steps should I use?
* What indicates the pipeline is done?
* What indicates that the upload is done?
  * Should I add console logs to these events to capture the timeline?

PR #1352, opened 2022-06-01
Vault: upload artifacts to s3

### LINKED ISSUE #1351: Vault: upload artifacts to s3
## Task Description
Vault runs without the user who is running the vault do not have access to files created on the vault server.

In a 'vault only run' with a computation that outputs file artifacts (vbm), those files are only available on the vault server. After a pipeline finishes these should be available (via the api) to anyone who has permissions to them.

## Task Notes
* change headless to vault in packaing and docs :), also change `cloud` to`vault` for naming sanity
* run a vault client run with VBM (stats) this will give you file output on finish
* on vault client finish (on a run with only vault clients??) upload everything in the `output/` dir to s3, using runid as a lookup key
* delete these artifacts locally on successful upload
* make an api call to store these in the db, perhaps in the run table?
* proxy the s3 download through hapi, there are some examples of this online (they might be express but it should be roughly the same), this puts access control to the s3 buckets in the hands of the api
* make a front end for this, likely somewhere w the `view results` button on the results page and widgets

### LINKED ISSUE #1366: UI: download and display s3 vault data
## Task Description
When a vault run has no normal users, the vault will upload the results to s3. The UI then needs to grab and unpack this for the user

## Task Notes
* download the resource based on the `runid` - this may not be available yet while it upload
* the `open local results` button should be `download results` if its a vault only run
* unpack the results as if there were normal results eg `COINSTAC_HOME/consortiumID/runID/{results}`
* the results page may need some work to not crash if the results are not available (before download)

dylanmartin:
- [x] Passes e2e testing
- [x] Docs have been added / updated if necessary (for bug fixes / features)

Feature

closes _issue #1358 #1351 #1350

When a pipeline runs with vault users only, one of the vault users will upload the run assets to s3 upon completion.

* **Other information**:
A next step will be to allow human users to download and view results in the COINSTAC ui


dylanmartin:
@rssk we still need to talk about changing "headless" to "vault" throughout the codebase.

dylanmartin:
https://github.com/trendscenter/coinstac/issues/1351#issuecomment-1165878983

packages/coinstac-api-server/src/routes/cloudFiles.js
```diff
@@ -0,0 +1,93 @@
...
+          const result = await downloadFromS3(fileKey);
+          return h.response(result.Body).code(200);
+        } catch (e) {
+          console.log(e);
+          return h.response(e).code(500);
+        }
+
+      
+      },
+      payload: {
```
dylanmartin, line 93: Try to upload and download a big file. 500mb

rssk:
Can you check if there are files to upload in the output directory?

packages/coinstac-client-core/src/index.js
```diff
@@ -435,6 +476,42 @@ class CoinstacClient {
...
+    axios.post(
+      `${process.env.API_URL}/uploadFiles`,
+      formData,
+      postConfig
+    ).then((result) => {
+      console.log(result);
+    }).catch((e) => {
+      console.log(e);
+    });
+
```
rssk: core has a logger instance that should be used for logs

packages/coinstac-ui/app/main/index.js
```diff
@@ -694,5 +699,100 @@ loadConfig()
...
+        });
+
+        // extract the tar
+        const readStream = fs.createReadStream(outputFilePath);
+        const writeStream = tar.extract(runOutputDirectory);
+        readStream.pipe(gunzip()).pipe(writeStream);
+
+        await new Promise((resolve, reject) => {
+          writeStream.on('finish', () => {
+            console.log('writeStream finished');
```
rssk: use the logger here too;)

packages/coinstac-ui/app/render/components/results/result.jsx
```diff
@@ -213,7 +224,36 @@ class Result extends Component {
...
+                  try {
+                    const result = await ipcRenderer.invoke('download-run-assets', {
+                      runId: run.id, authToken, clientId, apiServerUrl,
+                    });
+                    console.log(result);
+                    this.setState({ downloading: false });
+                    await this.doFilesExist(run.id);
+                  } catch (e) {
+                    console.log(e);
+                    this.setState({ downloading: false });
```
rssk: I'm guessing the top log is for debug, but the bottom should probably be wired to an error toast/notification

rssk (changes requested):

rssk (approved):

PR #1522, opened 2022-10-05
fix-1519

### LINKED ISSUE #1519: UI: Deal with when CSVs have columns with missing data elements
## Task Description
If a csv has rows with missing data, casting fails. It makes sense to catch these and remove them from the parsed csv

## Task Notes
* errors on casting are already caught, create a list of caught subjects from this
* remove those rows (this may be naturally happening via the catching)
* have the UI show a dialog with the removed subjects, extra credit: add the reason, this may not be useful to the user though or easy to figure out
* add this dialog above `Map CSV columns to pipeline variables` and below the show files dialog box
Screen Shot 2022-10-04 at 5 19 05 PM

dylanmartin:
- [x] Passes e2e testing

* **What kind of change does this PR introduce?**
Feature/fix

* **What is the current behavior?** 
closes _issue #1519


dylanmartin:
It would be nice if the check for bad rows happens before hitting the save button

dylanmartin:
image

dylanmartin:
The UX for this page could be redesigned based on the new information we're displaying. The redesign would be a bigger task. I think its worth it to ship the feature before redesigning since we're running into issues in production.

PR #1705, opened 2023-04-03
Fix/1694 compspec edit issue

### LINKED ISSUE #1694: Compspec edit needs revisit
Currently, the compspec shown to the user when clicking "GET IO" on a computation in the "computations" page is the result of some graphql query or subscription.

The return of that graphql function is a JSON object that contains a collection of keys and values. Currently, in that result, some necessary keys of the compspec are excluded (such as `compspecVersion`) , and some extraneous keys are included (such as the `delete` and `submittedBy` ). As a result:
* `comspecVersion` isn't shown, even if it is saved.
* The presence of the `delete` key when saving the compspec causes the action to be rejected.

If changing the return schema of the existing query or subscription is the to be considered as solution, it should be determined what other parts of code depend on the current return schema.

Another option is to create a new query designed specifically for compspec edit. That would avoid interfering with any code that consumes the return schemas for existing queries/subscriptions.

kalitopix:
- [] If your PR contains DB doc changes have you made an API migration and a localDB migration if necessary?
- [] Tests for the changes have been added (for bug fixes / features)
- [] Passes e2e testing
- [] Passes lint
- [] Docs have been added / updated if necessary (for bug fixes / features)

* **What kind of change does this PR introduce?**
Bug fix

* **What is the current behavior?** 
closes #1694 

* **Does this PR introduce a breaking change?**
No

### DIFF
--- a/packages/coinstac-desktop-app/app/render/components/computations/computation-io.jsx
+++ b/packages/coinstac-desktop-app/app/render/components/computations/computation-io.jsx
@@ -14,10 +14,24 @@ const styles = theme => ({
     marginTop: theme.spacing(1),
   },
   computationActions: {
+    marginTop: theme.spacing(1),
     display: 'flex',
     justifyContent: 'space-between',
   },
 });
+
+const prepareJSON = (info) => {
+  const compSpecDbObject = JSON.parse(JSON.stringify(info));
+
+  delete compSpecDbObject.id;
+  delete compSpecDbObject.__typename;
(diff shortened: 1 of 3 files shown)


kalitopix:
@dylanmartin Updated graphql schema to include `compspecVersion` and exclude `delete`.
API sends `submittedBy` but it's not being sent whe compspec is updated.

rssk (approved):

Issue #1779, opened 2023-07-23
workshop bugs: Add Green tick for "Save pipeline" 


spanta28:
After adding pipeline in a consortia, when user clicks "Save pipeline", add a green tick mark to notify user that the pipeline is saved.

Issue #1774, opened 2023-07-23
workshop bugs: Set pipeline active on consortia by default


spanta28:
Set pipeline active on consortia by default

<img width="640" alt="Screenshot 2023-07-23 at 5 26 09 PM" src="https://github.com/trendscenter/coinstac/assets/5415931/b4d29a4b-38d3-4bc5-891f-d5b3aecf9a7d">

spanta28:
Confirmed today that we can assume any new pipeline that is created will be the one that will be active. So we can remove the "Set active pipeline" option and make this default behaviour. Users can change to other pipelines created in the consortia by going to "Pipelines" page and selecting the pipeline from "Owned Pipelines"

PR #1804, opened 2023-08-09
Setting Pipeline active in current Consortium by default

pixelsaurus:
See #1774 and/or #1779


rssk:
this appears to be a duplicate PR? 
https://github.com/trendscenter/coinstac/pull/1798

Issue #1833, opened 2023-10-05
Populate ROI's in real time for Freesurfer computations


spanta28:
For the foll. computations, in the pipeline setup page, we want to populate ROI's by having the user upload a sample Freesurfer stats file (*aseg*stats* from fs outputs). Example file here: 
[subject0_aseg_stats.txt](https://github.com/trendscenter/coinstac/files/12824179/subject0_aseg_stats.txt)

https://github.com/trendscenter/coinstac-regression-fsl
https://github.com/trendscenter/coinstac-ssr-fsl
https://github.com/trendscenter/coinstac-msr-fsl

PR #1841, opened 2023-10-17
Feature/pipeline populate ro is from file 1833

pixelsaurus:
See #1833 

Also fixed issue with cleared ROIs return a breaking null error


pixelsaurus:
Didn't have updated file from master. Need to refactor...

